PR #1, opened 2021-10-01
Added WorldManager and StateManager

Gursaanj:
Create Base Structure for the WorldManager and StateManager (in the Manager Scene). No real logic implemented as of yet!


Assets/_BForBoss/_LevelDesign/Managers/StateManager.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+
+namespace BForBoss
+{
+    public enum State
+    {
+        PreGame,
+        Play,
+        Pause,
+        Death,
```
Gursaanj, line 10: This seems appropriate for now, but I would really like to push for a more robust Statemachine in the future with half step states (i.e. how Unity handles it[ PlayState ](https://docs.unity3d.com/ScriptReference/PlayModeStateChange.html). This is so that we can handle things like Cleanup, Reset and other similar management in the in between states and leave the main states with less clutter work to do

m4rtint: yeah for sure, that definitely would be ideal. 

Pros of that - more flexible and fine tuned
Cons of that - it can get bloated and way too much. 

So might need to have sub states as well, or like create a separate life cycle for the substate e.g. the PlayState. Just something to think about

Assets/_BForBoss/Scripts/Editor/HierarchyWindowGroupHeader.cs
```diff
@@ -0,0 +1,35 @@
+//-----------------------------------------------------------------------
+// <copyright file="HierarchyWindowGroupHeader.cs" company="PERIGON GAMES">
+//     Copyright (c) Perigon Games. 2020 All Rights Reserved
+// </copyright>
+//-----------------------------------------------------------------------
+using UnityEditor;
```
m4rtint, line 6: We don't need this, It's already part of the project under our PerigonTools section

Assets/_BForBoss/_LevelDesign/Managers/StateManager.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+    
+    public class StateManager
```
m4rtint, line 13: This is great, only thing missing, is an interface where you can getState and SetState().

So that you can inject a MockStateManager into any manager/component that needs it.

Assets/_BForBoss/_LevelDesign/Managers/WorldManager.cs
```diff
@@ -0,0 +1,59 @@
+using UnityEngine;
+
+namespace BForBoss
+{
+    public class WorldManager : MonoBehaviour
+    {
+        private StateManager _stateManagerInstance = StateManager.Instance;
```
m4rtint, line 7: NIT: probably don't need to name it `_stateManagerInstance `
just name it `_stateManager`

Assets/_BForBoss/Scripts/Editor/HierarchyWindowGroupHeader.cs
Gursaanj: Should I Import PerigonTools into this project in this PR then? or just delete this file for now?

m4rtint: that's strange, does it not already have perigon tools? it should be in ur manifest file. 

But try to delete this file, and see if it still works

Gursaanj: My bad, it was there

m4rtint (approved):

PR #7, opened 2021-10-09
BFB-38: Integrate Player Prefab into the racecourse

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-38

**Description**
- Added player prefab into the race course


unoctium1:
One thing to note is that crouching doesn't work here - the cinemachine brain needs its default blend time reduced in order to work with the fps controller. I think for now we should leave this as is so you're dolly track isn't broken, and I'll come up with a more graceful way of handling this in the first person controller

unoctium1 (approved):

m4rtint:
> One thing to note is that crouching doesn't work here - the cinemachine brain needs its default blend time reduced in order to work with the fps controller. I think for now we should leave this as is so you're dolly track isn't broken, and I'll come up with a more graceful way of handling this in the first person controller

That's okay for now, since i'm just focusing on the movement and jump. Don't need to focus on the crouch just yet until slide.

PR #9, opened 2021-10-12
Martin/git ignore update

m4rtint:
**Description**
- Removed unitypackage.meta 
- Removed EditorUserSettings - it's generated as per your own local settings
- https://forum.unity.com/threads/whats-the-usersettings-directory.754436/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,4 +65,4 @@ ProjectSettings/Packages/com.unity.progrids/Settings.json
 Assets/ECM2/Samples/*
 UserSettings/EditorUserSettings.asset
 .DS_Store
-.unitypackage.meta
+*.unitypackage.meta


.gitignore
```diff
@@ -63,3 +63,6 @@ ProjectSettings/Packages/com.unity.probuilder/Settings.json
 ProjectSettings/Packages/com.unity.progrids/Settings.json
 .idea/
 Assets/ECM2/Samples/*
+UserSettings/EditorUserSettings.asset
+.DS_Store
+.unitypackage.meta
```
unoctium1, line 68: Should this be *.unitypackage.meta?

unoctium1: Huh, I always thought it was some modified version of regex too haha

PR #15, opened 2021-10-15
BFB-32: Character Sliding

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-32

**Description**
*NOTE - I realized that I rather allow flexible abilities of anything goes, and then tune it down as we iterate
- Able to slide
- Slide any direction
- won't be able to Slide on stand still

**Screenshots / GIFs**
https://youtu.be/vlwCGmOL680


Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -46,16 +53,54 @@ protected override void AnimateEye()
...
+
+        public override float GetMaxSpeed()
+        {
+            if (_slideBehaviour != null && _slideBehaviour.IsSliding)
+            {
+                return _slideBehaviour.MaxWalkSpeedSliding;
+            }
+
+            return base.GetMaxSpeed();
+        }
```
Gursaanj, line 74: public should be above if Im not mistaken

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -46,16 +53,54 @@ protected override void AnimateEye()
...
+
+        protected override Vector3 CalcDesiredVelocity()
+        {
+            if (_slideBehaviour != null && _slideBehaviour.IsSliding)
```
Gursaanj, line 78: Up to you but it could be cleaner to just make a property for this bool (since it being used a lot)

i.e. `private bool IsSliding => _slidebehaviour != null && _slidebehaviour.IsSliding`.

Similarly you could make a method to check

Gursaanj: then you could position all these methods with ternary operators. i.e. `returning IsSliding ? Vector3.zero " Base.CalcDesiredVelocity()`

Gursaanj (commented):
Small Comments

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
m4rtint: Moved it.

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
m4rtint: Made the. changes

Gursaanj (approved):

PR #17, opened 2021-10-16
BFB-36: Speedometer for character

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-36

**Description**
- Create Speedometer for character

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/137598727-bfdf152a-a144-4495-a3f2-09499afa4b47.mp4


Assets/_BForBoss/Scripts/SpeedometerBehaviour.cs
```diff
@@ -0,0 +1,55 @@
...
+        float Speed { get; }
+    }
+    
+    public class SpeedometerBehaviour : MonoBehaviour
+    {
+        [SerializeField] private Image _meter = null;
+        [SerializeField] private TMP_Text _speedLabel = null;
+        
+        private ICharacterSpeed _characterSpeed = null;
+        private float _maxSpeed = 50f;
```
Gursaanj, line 18: Should this have the `[SerializedField]` attribute on it so it's easier to edit?

Assets/_BForBoss/Scripts/SpeedometerBehaviour.cs
```diff
@@ -0,0 +1,55 @@
...
+                return;
+            }
+            
+            SetMeter();
+            SetSpeedLabel();
+        }
+
+        private void SetMeter()
+        {
+            var percentage = _characterSpeed.Speed / _maxSpeed;
```
Gursaanj, line 46: if you do end up setting the _maxSpeed variable to be editable via insepctor, make sure it has the `[Min]` attribute (set to at least 1f) so that this value is never 0 and cause an exception

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -6,16 +6,15 @@
 
 namespace BForBoss
 {
-    public class FirstPersonPlayer : FirstPersonCharacter
+    public partial class FirstPersonPlayer : FirstPersonCharacter
```
Gursaanj, line 9: What is the reason for using a partial class and not just allowing `FirstPersonPlayer` to inherit from `ICharacterSpeed`?

Gursaanj (commented):
Small comments and question

Assets/_BForBoss/Scripts/SpeedometerBehaviour.cs
m4rtint: yeah probably. this is sort of a placeholder.

Assets/_BForBoss/Scripts/SpeedometerBehaviour.cs
m4rtint: yup for sure

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
m4rtint: So this class here is our base `FirstPersonPlayer` that does what it needs. 
So if we copy this class to another project, it'll work as intended. 

And then underneath (the other partial), are all the extra methods that it needs to implement so I could inject it into other modules else where i.e. Whatever UI that needs specific data from this class.

So if we were to move this class to FNF, we only need to move this partial class, without the extra `ICharacterSpeed` stuff. 

hope this makes sense.

m4rtint: Ideally, the partial class: `ICharacterSpeed` would be in another file. But I was just lazy.

Gursaanj (approved):

PR #19, opened 2021-10-17
BFB-35: Base Wall Running Mechanic

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-35

**Description**
Create basic wall running mechanic
Currently includes a configurable speed multiplier, configurable minimum height off the ground, maximum distance from the wall, and a (basic) gravity implementation. Also includes a camera roll effect, though this can be taken out if need be

**Screenshots / GIFs**
[Video Demo](https://www.youtube.com/watch?v=LqH3CZglewE)


Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+        public bool CanJump()
+        {
+            return _isWallRunning; // can always jump out of a wall run
+        }
+
+        public void OnWallRunning()
+        {
+            if (!_isWallRunning)
+            {
+                _timeSinceWallDetach += Time.deltaTime;
```
Gursaanj, line 117: more of a question, but should we be using `Time.FixedDeltaTime` instead since we are dealing with time in relation to movement operations?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
+using ECM2.Characters;
+using ECM2.Components;
+using ECM2.Common;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
```
Gursaanj, line 7: Is this being used somewhere?

Gursaanj (commented):
The linked video is private,  wasn't able to open it!

Looks good, will look over again in the morning

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -15,10 +15,19 @@ public class FirstPersonPlayer : FirstPersonCharacter
...
+        {
+            if (_wallRunBehaviour != null)
+                return _wallRunBehaviour.CanJump() || base.CanJump();
+            return base.CanJump();
+        }
 
         protected override void OnAwake()
         {            
             _dashBehaviour = GetComponent<PlayerDashBehaviour>();
+            _wallRunBehaviour = GetComponent<PlayerWallRunBehaviour>();
```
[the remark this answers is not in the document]
m4rtint: we'll see :) .but if it's just this one line, i think it'll be okay.

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            _mask = LayerMask.GetMask("ParkourWall");
+        }
+
+        public void Falling(Vector3 _)
+        {
+            if (!CanWallRun()) return;
+            _hits = new RaycastHit[directions.Length];
+            for (int i = 0; i < directions.Length; i++)
+            {
+                Vector3 dir = ChildTransform.TransformDirection(directions[i]); //convert directions to relative to player
```
m4rtint, line 76: can you remove the comment and make this variable more verbose? 
e.g. `playerFacingDirection`

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            float cameraAngle = _baseCharacter.cameraTransform.eulerAngles.z;
+            float targetAngle = 0;
+            if (dir != 0)
+            {
+                targetAngle = Mathf.Sign(dir) * _maxCameraAngleRoll;
+            }
+            return Mathf.LerpAngle(cameraAngle, targetAngle, Mathf.Max(_timeSinceWallAttach, _timeSinceWallDetach) / _cameraRotateDuration);
+        }
+
+        private static bool GetSmallestRaycastHit(RaycastHit[] array, out RaycastHit smallest)
```
m4rtint, line 210: Rename this to `GetSmallestRaycastHitIfValid`

so when placed into an if statement, it is more sound.

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            _timeSinceWallAttach = 0f;
+            _timeSinceWallDetach = 0f;
+        }
+
+        private float CalculateSide()
+        {
+            if (_isWallRunning)
+            {
+                Vector3 heading = _lastWallRunPosition - transform.position;
+                Vector3 perp = Vector3.Cross(ChildTransform.forward, heading);
```
m4rtint, line 191: can you use full word? I think this is perpendicular?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+        private float _minHeight = 1f;
+        [SerializeField]
+        private float _wallMaxDistance;
+        [SerializeField]
+        private float _wallGravityDownForce = 0f;
+        [SerializeField, Range(0f, 3f), Tooltip("Only allow for a wall run if jump is longer than this")]
+        private float _minJumpDuration = 0.3f;
+
+        [SerializeField]
+        private float _maxCameraAngleRoll;
```
m4rtint, line 27: Can you set default values for these?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+        private void StopWallRunning()
+        {
+            if (!_isWallRunning) return;
+            _isWallRunning = false;
+            _baseCharacter.maxWalkSpeed = _baseMaxSpeed;
+            _timeSinceWallAttach = 0f;
+            _timeSinceWallDetach = 0f;
+        }
+
+        private float CalculateSide()
```
m4rtint, line 186: I'm assuming the `side` you're referring to, is the side of the wall relative to the player?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            {
+                Vector3 heading = _lastWallRunPosition - transform.position;
+                Vector3 perp = Vector3.Cross(ChildTransform.forward, heading);
+                float dir = Vector3.Dot(perp, ChildTransform.up);
+                return dir;
+            }
+            return 0;
+        }
+
+        private float GetCameraRoll()
```
m4rtint, line 198: Can you place all methods related to Camera Rolling under a #Region?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+        }
+
+        private bool CanWallRun()
+        {
+            if (_baseCharacter.IsOnGround()) return false;
+
+            if (_baseCharacter.IsJumping())
+            {
+                _currentJumpDuration += Time.deltaTime;
+                if (_currentJumpDuration < _minJumpDuration) return false;
```
m4rtint, line 171: Out of curiosity, what happens if you don't have this check? Why do you need to check this?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            }
+            _timeSinceWallAttach += Time.deltaTime;
+            var velocity = _baseCharacter.GetVelocity();
+            var alongWall = ChildTransform.TransformDirection(Vector3.forward).normalized;
+            velocity = velocity.dot(alongWall) * alongWall;
+            velocity += Vector3.down * _wallGravityDownForce * Time.deltaTime;
+            _baseCharacter.SetVelocity(velocity);
+        }
+
+        public bool CalcJumpVelocity(out Vector3 velocity)
```
m4rtint, line 133: Why do you need an `out ` here? why not just create a new Vector3 and return that, since it starts at Vector3.zero anyway?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+        private bool CanWallRun()
+        {
+            if (_baseCharacter.IsOnGround()) return false;
+
+            if (_baseCharacter.IsJumping())
+            {
+                _currentJumpDuration += Time.deltaTime;
+                if (_currentJumpDuration < _minJumpDuration) return false;
+            }
+
```
m4rtint, line 173: how hard is it to check that the player has a forward momentum? Basically only allow wall running if the player is pressing {Forward Button} + {Left Button} || {Right Button}.

Since currently you can wall run backwards as well

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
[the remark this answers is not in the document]
unoctium1: Yeah haha I'm not optimistic, I'll get that all merged in and sorted out though

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: Good catch, I had that in for debug stuff and forgot about it

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: Sure, I'll go through and wrap everything else in regions too

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: You can't jump out of a wall run without it, since you essentially stick to the wall

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+            velocity += Vector3.down * _wallGravityDownForce * Time.deltaTime;
+            _baseCharacter.SetVelocity(velocity);
+        }
+
+        public bool CalcJumpVelocity(out Vector3 velocity)
+        {
+            velocity = Vector3.zero;
+            if (_isWallRunning)
+            {
+                velocity = _lastWallRunNormal * _baseCharacter.jumpImpulse + Vector3.up;
```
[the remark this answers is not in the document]
unoctium1: Oh good idea. I could also multiply just the up movement by jump impulse and have a separate tuning variable for the out of the wall jump, what do you think?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: Its how I was overriding things in the first person controller, since I only want to use the wallrunbehaviour's jump velocity if we're wall running. Using an out saves a vector comparison

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: Hmm I think you might be right, I'll update those

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -0,0 +1,232 @@
...
+
+            if (_movementInput().sqrMagnitude <= 0)
+            {
+                return;
+            }
+            _timeSinceWallAttach += Time.deltaTime;
+            var velocity = _baseCharacter.GetVelocity();
+            var alongWall = ChildTransform.TransformDirection(Vector3.forward).normalized;
+            velocity = velocity.dot(alongWall) * alongWall;
+            velocity += Vector3.down * _wallGravityDownForce * Time.deltaTime;
```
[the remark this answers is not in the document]
unoctium1: 127 and 128 are just base movement, its just 129 thats gravity over time. I'm going to reset the time factor to Time.fixedUpdate per Gursaanj's comment, but I think in general if we're applying a force we don't want it to be a constant force per frame - that would require very low and hard to tune variables

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: Yeah I think if I just call the input func and check if the y component is positive that should work, I'll test it

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: It's just personal general coding habits, by default I set it to non-static. Looking through the method it seems fine in this case. 
and just read through when other's set to static.
https://stackoverflow.com/questions/2713601/when-to-make-a-method-static

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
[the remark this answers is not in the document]
m4rtint: We'll might need to tune this over time, we can change it to separate values if we think it's not strong enough

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: oh i see.

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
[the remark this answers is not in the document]
m4rtint: can you set `Vector3.down * _wallGravityDownForce * Time.deltaTime` to be a separate var called `dragDownVelocity`

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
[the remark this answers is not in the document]
Gursaanj: yeah we will probs have own our own Multiplier variable we will be multiplying by

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: yeah alright, i think it keeps it consistent

unoctium1:
> The linked video is private, wasn't able to open it!
> 
> Looks good, will look over again in the morning

Video should be fixed now!

m4rtint:
> I have realised that dashing and wall running are possible at the same time. However, it plays okay in my opinion, not sure if we actually need to prevent that

I think it should be okay.

m4rtint (approved):
This is a great start :)

PR #21, opened 2021-10-17
BFB-46: Dash Lines VFX

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-46

**Description**
- Added Particle Effects going outwards - inspired by Valorant.
- *NOTE* - If we were using Doom Inspiration, we would have to use Shader/Shader Graph, and Particle Effects seems more straightforward for this iteration. 

- Lens distortion to make it look fast - I also looked at Motion Blur + Depth of Field, but they don't blur the edges as i wanted it to.

- Added another Post Processing volume (Lens Distortion) to change the weight and show it during dash.

**Screenshots / GIFs**
https://www.youtube.com/watch?v=-Toe6KMDVX0

Screen Shot 2021-10-17 at 4 21 12 PM


Assets/_BForBoss/_Character/Scripts/LensDistortionTool.cs
```diff
@@ -0,0 +1,35 @@
+using DG.Tweening;
+using UnityEngine.Rendering.Universal;
+
+namespace BForBoss
+{
+    /// <summary>
+    /// Tool to distort the lens.
+    /// WARNING/NOTE - Post Processing > Volume > LensDistortion - make sure intensity is turned on
+    /// For some reason there's no built in function to enable this automatically.
```
m4rtint, line 9: Note*
You MUST have LensDistortion > Intensity turned manually in inspector on for this to work.

m4rtint:
> It might be more straightforward and extensible to just assign the lens distortion to a custom post processing volume, and tween the volume weight up in a dash instead of trying to manually alter the specific effect itself

Updated to use weights
1. Added another Global Post Processing Volume
2. Changed the Tool to Tween the weight for the volumes

Screen Shot 2021-10-17 at 4 21 12 PM
.

unoctium1 (changes requested):
The effect looks great, and plays well in the scene. Only requesting changes cause it broke when merging in wall running, I added the dash component to the prefab and so now its grabbing the wrong component. I'm happy to fix this on my end and push the changes

PR #24, opened 2021-10-18
BFB-37: Wall Run Timer, and other wall run tweaks

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-37

**Description**
Add timer to wall run gravity. Can reopen if we feel like distance is better, but timer should be adequate. 

Also add velocity to jumping out of a wall run,  jump height multiplier, and override to max acceleration


m4rtint (approved):
this is really great . Great job :)

PR #45, opened 2021-10-30
BFB-62: Analytics for checkpoints

mikeslgoh:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-62

**Description**
- fire an event whenever a player crosses a new checkpoint
- had to list the field as timer because Mixpanel was replacing the time field with their own field (as seen in the screenshot, there is an extra field)

**Screenshots / GIFs**
image

### DIFF
--- a/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
+++ b/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
@@ -76,7 +76,7 @@ namespace BForBoss
             Mixpanel.Track(Event.PlayerDeath, props);
         }
 
-        public void LogCheckpointData(float time, String checkpointName)
+        public void LogCheckpointEvent(float time, String checkpointName)
         {
             String formattedTime = string.Format($"{time:0.00}s");
             var props = new Value();
--- a/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
+++ b/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
@@ -78,10 +78,9 @@ namespace BForBoss
 
         public void LogCheckpointEvent(float time, String checkpointName)
         {
-            String formattedTime = string.Format($"{time:0.00}s");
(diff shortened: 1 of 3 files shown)


Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
```diff
@@ -57,16 +64,31 @@ public void EndSession()
...
             
             props[EventAttribute.Course] = EventConstant.RaceCourse;
-            props[EventAttribute.Name] = name;
+            props[EventAttribute.Name] = deathAreaName;
             Mixpanel.Track(Event.PlayerDeath, props);
         }
 
+        public void LogCheckpointEvent(float time, String checkpointName)
+        {
+            String formattedTime = string.Format($"{time:0.00}s");
```
m4rtint, line 81: hmm sorry. now that i think about it, can you set the time to be a float? 

That way we can actually chart it.

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
```diff
@@ -4,21 +4,25 @@
...
     {
         public const String PlayerDeath = "Player - Death";
+        public const String CheckpointReached = "Checkpoint - Reached";
     }
 
     public readonly struct EventAttribute
     {
         public const String Course = "course";
         public const String Name = "name";
+        public const String Time = "timer";
```
m4rtint, line 18: can you name it "time"

mikeslgoh: I can't name it to time because when I used time, mixpanel overwrites it for some reason. Was not sure what else to change it to. I could try removing the sessions?

m4rtint: ah that's unfortunate. that's okay then.

m4rtint (approved):
lgtm

PR #47, opened 2021-10-30
BFB-64: Set username property

mikeslgoh:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-64

**Description**
- uses the username saved in Playerprefs (if already exists)
- if not, will set the username once the player clicks save on the panel

**Screenshots / GIFs**
image

### DIFF
--- a/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
+++ b/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections;
 using mixpanel;
-using UnityEngine;
 
 namespace BForBoss
 {
@@ -27,7 +26,7 @@ namespace BForBoss
 
     public interface IPerigonAnalytics
     {
-        void StartSession();
+        void StartSession(String uniqueId);
         void EndSession();
         void LogDeathEvent(String name);
         void LogEvent(String eventName);
@@ -60,9 +59,9 @@ namespace BForBoss
         
         #endregion
         
-        public void StartSession()
+        public void StartSession(String uniqueId)
         {
-            Mixpanel.Identify(SystemInfo.deviceUniqueIdentifier);
+            Mixpanel.Identify(uniqueId);
             Mixpanel.Track(SessionStart);
         }
 
--- a/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
+++ b/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using mixpanel;
+using UnityEngine;
 
 namespace BForBoss
 {
@@ -54,7 +55,6 @@ namespace BForBoss
 
         public void SetSessionUsername(String username)
         {
-            Mixpanel.Identify(username);
             Mixpanel.People.Set("$name", username);
         }
         
@@ -62,6 +62,7 @@ namespace BForBoss
         
         public void StartSession()
         {
+            Mixpanel.Identify(SystemInfo.deviceUniqueIdentifier);
             Mixpanel.Track(SessionStart);
         }
 
--- a/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
+++ b/Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
@@ -43,13 +43,19 @@ namespace BForBoss
 
         public static PerigonAnalytics Instance => _instance;
 
-        #region CONSTRUCTORS
+        #region CONSTRUCTORS / SETTERS
         static PerigonAnalytics()
         {
         }
 
         private PerigonAnalytics()
         {
+        }
+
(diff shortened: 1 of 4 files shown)


mikeslgoh:
Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel

m4rtint:
> Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel

It shouldny be an event, it should be a user property

m4rtint:
> Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel

and yes, player is able to change their name

mikeslgoh:
> > Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel
> 
> It shouldny be an event, it should be a user property

yup it isn't an event

m4rtint:
> > > Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel
> > 
> > 
> > It shouldny be an event, it should be a user property
> 
> yup it isn't an event

>when a unique ID is first created for a user (e.g., when a user first registers for an account)
https://developer.mixpanel.com/docs/unity

Because we don't have have auth flow, u can use something like HWID as a unique identifier or something. Hopefully it won't be super detrimental in the future

mikeslgoh:
> > > > Question @m4rtint : are players going to be able to change their username? if so, I'll need to save a distinct id in player prefs so that the user is free to use whatever name they want but a new profile isn't created on mixpanel
> > > 
> > > 
> > > It shouldny be an event, it should be a user property
> > 
> > 
> > yup it isn't an event
> 
> > when a unique ID is first created for a user (e.g., when a user first registers for an account)
> > https://developer.mixpanel.com/docs/unity
> 
> Because we don't have have auth flow, u can use something like HWID as a unique identifier or something. Hopefully it won't be super detrimental in the future

yeah I'll try that and see. shouldn't be an issue?

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using mixpanel;
+using UnityEngine;
```
m4rtint, line 4: Is this needed?

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
```diff
@@ -43,19 +44,25 @@ public class PerigonAnalytics : IPerigonAnalytics
...
         {
         }
 
         private PerigonAnalytics()
         {
         }
+
+        public void SetUsername(String username)
+        {
+            Mixpanel.People.Set("$name", username);
```
m4rtint, line 58: Is the $ so that it overrides mixpanel default name

Assets/_BForBoss/_Leaderboard/Scripts/InputUsernameViewBehaviour.cs
```diff
@@ -32,6 +33,7 @@ public void Initialize(InputUsernameViewModel viewModel)
             _saveUsernameButton.onClick.AddListener(() =>
             {
                 _viewModel.SetUserName(_usernameField.text);
```
m4rtint, line 35: I'm wondering can you `MixPaanel.setUserName` within the viewModel. since viewModels are more reused than View's

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
mikeslgoh: the $ is just part of the field name for mixpanel so that I can set its value

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
mikeslgoh: I needed this in order to use the device unique identifier. If we don't want the UnityEngine import, I can try looking for another unique ID

m4rtint (approved):
Lgtm

Assets/_BForBoss/_Analytics/Scripts/PerigonAnalytics.cs
m4rtint: technically you can inject the string instead. So that it's not couple to unity engine

m4rtint: I'll let you decide

mikeslgoh: yup that's probably better. I made the changes

PR #48, opened 2021-11-01
BFB-65: Set controller settings property

mikeslgoh:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-65

**Description**
- set control settings for mouse and controller in InputSettings view model
- not sure where else I should set the control settings property
- also a bug fix where if the user was deleted from mixpanel BUT was already set in PlayerPrefs, a new user would be created but the username would need be saved to the mixpanel profile

**Screenshots / GIFs**
image

### DIFF
--- a/Assets/_BForBoss/Scripts/InputSettingsViewBehaviour.cs
+++ b/Assets/_BForBoss/Scripts/InputSettingsViewBehaviour.cs
@@ -105,7 +105,8 @@ namespace BForBoss
     public class InputSettingsViewModel
     {
         private readonly IInputSettings _settings = null;
-        
+        private readonly PerigonAnalytics _perigonAnalytics = PerigonAnalytics.Instance;
+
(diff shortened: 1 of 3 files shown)


Assets/_BForBoss/Scripts/InputSettingsViewBehaviour.cs
```diff
@@ -141,6 +144,18 @@ public void ApplySettings(float horizontalMouse, float verticalMouse, float hori
             _settings.ControllerHorizontalSensitivity = horizontalController;
             _settings.ControllerVerticalSensitivity = verticalController;
             _settings.IsInverted = isInverted;
+
+            SetControlSettings();
+        }
+
+        public void SetControlSettings()
```
m4rtint, line 151: Does this need to be public?

m4rtint (approved):
lgtm

PR #49, opened 2021-11-01
Bfb 40

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-40

**Description**
Added Development_Build friendly DebugWindow for all our Future Debug options that we would like to have access to during the editor and the build. Also added is the SceneSwitcher DebugOption. Which lets the player switch between RaceCourse, TrainingCourse, and SandBox Scene!

To open DebugWindow Click LeftAlt + F1 (let me know if binding needs to change for Mac Users). Then Click the SceneSwithcer Option and select which scene you would like to test.

https://youtu.be/4M9o3J8jh58

There is one issue that comes with this. Though this does not break implementation at all it seems that the Unity Input System Still tries to access The FirstPlayerScript (when trying to click on screen or Press Escape) even when the Scene has Changed causing some MissingReferenceExceptions!. Still looking into this for now!

Unwanted Input Error


Assets/_BForBoss/Scripts/Managers/WorldManager.cs
```diff
@@ -96,10 +111,16 @@ private void HandleStateChange(State newState)
                 }
                 case State.Play:
                 {
+                    if (Time.timeScale != 1.0f)
```
m4rtint, line 114: Do you need this check?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -8,6 +8,10 @@ namespace BForBoss
 {
     public partial class FirstPersonPlayer : FirstPersonCharacter
     {
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
```
m4rtint, line 11: The FIrstPersonPlayer should not have any changes. Since now, this script is tightly coupled with the editor tool.

Is it crucial that the actions needs to be disabled?

Assets/_BForBoss/Scripts/Managers/WorldManager.cs
```diff
@@ -19,6 +19,11 @@ public class WorldManager : MonoBehaviour
 
         [Title("Effects")] 
         [SerializeField] private Volume _deathVolume = null;
+        
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+        [Title("Debug")]
+        [SerializeField] private GameObject _debugCanvas;
```
m4rtint, line 25: wondering why this has to be in WorldManager, why can't this be drag and drop into a scene?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -66,6 +86,34 @@ protected override void OnStart()
...
+
+        protected override Vector2 GetMouseLookInput()
+        {
+            if (mouseLookInputAction != null && !IsDebugWindowOpen)
+                return mouseLookInputAction.ReadValue<Vector2>();
+
+            return Vector2.zero;
+        }
+
+        protected override void OnCursorLock(InputAction.CallbackContext context)
```
m4rtint, line 99: do you still need this? since you're using the mouse utility?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -66,6 +86,34 @@ protected override void OnStart()
             _wallRunBehaviour?.Initialize(this, base.GetMovementInput, ResetJumpCount);
         }
 
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+
+        protected override Vector2 GetMouseLookInput()
```
m4rtint, line 91:
```suggestion
                protected override Vector2 GetMouseLookInput()
        {
            if (IsDebugWindowOpen)
            {
                return Vector2.zero;
            }
            
            return base.GetMouseLookInput();
        }

        protected override void OnCursorLock(InputAction.CallbackContext context)
        {
            if (IsDebugWindowOpen)
            {
                return;
            }
            base.OnCursorLock(context);
        }
```

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -20,6 +24,22 @@ public partial class FirstPersonPlayer : FirstPersonCharacter
         public void Initialize()
         {
             SetupInput();
+            
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+            UnityEngine.SceneManagement.SceneManager.sceneUnloaded += arg0 => 
+            {
+                movementInputAction?.Disable();
```
m4rtint, line 31: Can you put all these `Disable` into the `OnOnDestroy` method instead?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
m4rtint: There's also another issue here. 
The character shouldn't know about this so called `DebugWindow`, 
If thet pause screen shows up, it shouldn't know about the "pause screen" either.

What it should know, is when to `freeze` or `StopAllActions`.  And it shouldn't be behind a #Debug
This is more of a naming issue

Assets/_BForBoss/Scripts/Debug/DebugWindow.cs
```diff
@@ -0,0 +1,172 @@
...
+                _currentDebugView.ResetData();
+                _currentDebugView = null;
+            }
+        }
+
+        private void OpenPanel()
+        {
+            _currentState = _stateManager.GetState();
+            _stateManager.SetState(State.Pause);
+            FirstPersonPlayer.IsDebugWindowOpen = true;
```
m4rtint, line 135:
```suggestion
            IInputSettings input = FindObjectOfType<FirstPersonPlayer>();
            input.DisableActions();
```

Disabling/Enabling the Action within IInputSettings might be more cleaner, since we'll be needing it in Other settings menu

m4rtint: Inside IInputSettings
```suggestion
                   public void DisableActions()
        {
            controllerLookInputAction.Disable();
            mouseLookInputAction.Disable();
        }

        public void EnableActions()
        {
            controllerLookInputAction.Enable();
            mouseLookInputAction.Enable();
        }
```

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
m4rtint: Nevermind, i understand what this is for now. 

Unfortunately, i can't seem to find a way around this. Best way to handle this is to place this override method within `partial FirstPersonPlayer.Debug class`

And then set the "Debug Flag" on over there.
```
public partial FirstPersonPlayer {
  //override code
}

```

Assets/_BForBoss/Scripts/Debug/DebugWindow.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Reflection;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+namespace BForBoss
+{
+    public class DebugWindow : MonoBehaviour
+    {
+
+        private const Key _keyCodeCharacter = Key.F1;
```
m4rtint, line 12:
```suggestion
        private const Key _keyCodeCharacter = Key.Backquote;
        private const Key _keyCodeModifier = Key.LeftCtrl;
```

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
mikeslgoh: Oh I did not know you could do suggested changes O_o

Assets/_BForBoss/Scripts/Managers/WorldManager.cs
Gursaanj: Yeah I was thinking of not adding the check There is no way that State.PreGame or that State.EndRace will have a different timescale?

Gursaanj: I can remove this for now

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
Gursaanj: Fair point. So would you rather let DebugWindow have an instance of FirstPersonPlayer?

Gursaanj: Unfortunately, we do have to Disable this, cause mouse clicking over anything not a gameobject (i.e. GUI layout buttons) will force the Cursor to lock

Assets/_BForBoss/Scripts/Managers/WorldManager.cs
Gursaanj: My thought Process is that we should'nt need to remember to delete the DebugWindow Prefab before we make a production Build. Like, I want it to "automate" so that we have to remember to delete/add things to the scene before we make a build

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
Gursaanj: OOh I like the Partial Class Idea. Yeah I was scratching my head on this for a while to be honest, really did not want to add this to the FirstPersonPlayer Class!.

Gursaanj: The reason I would rather use GuiLayout that UGUI is for scalability. The time and effort needed to create UGUI everytime you want to create a Debug Option is a bit much in my opinion. Plus a lot more flexibility with GUI I feel

Assets/_BForBoss/Scripts/Managers/WorldManager.cs
m4rtint: i agree with the reminder part. However, WorldManager can change quite often. 
Even possible that we don't use this worldmanager for certain "scenes/worlds". 

Shouldn't we leave this in, even when building the prod build? since it won't be compiled by prod anyway? 
And do you think it's better to have the debug stuff handle all this by itself instead of linking it to worldmanager?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
Gursaanj: Actually Im planning to remove this altogether. this was my original idea to solve the Error Message I posted in the description, but was only working some of the times. Also I think `OnDisable` is called when `OnDestroy` is called. Will really have to look into this soon

Assets/_BForBoss/Scripts/Debug/DebugWindow.cs
Gursaanj: Oh good point, made a similar script as Im finishing the Pause menu haha. Should just use this instead haha

Assets/_BForBoss/Scripts/Debug/DebugWindow.cs
Gursaanj: The thing is with LeftCTrl is that the player will crouch before the time Pauses. This might not be wanted by the user? What do you think?

m4rtint: so 2 options
1. it's for explicitly for the devs, so UX is not my top priority (and its like the only option between Mac and Windows which is common and not super weird)
2. just use the Backquote - ` key.

Assets/_BForBoss/Scripts/Managers/WorldManager.cs
Gursaanj: hmm good point, we can discuss this during the meeting today

Assets/_BForBoss/Scripts/Debug/DebugWindow.cs
Gursaanj: hmm, we can also create a command specifically for Mac and then use `#If Unity_MAC` pragma, if you want to use the modifier/character combo

Gursaanj: Im also okay using backquote too for now

m4rtint: just easier to backquote. but it's ultimately up to you.

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
Gursaanj: So now this is handled strictly via `IInputSettings`

m4rtint (approved):
lgtm

PR #54, opened 2021-11-07
BFB-87: Settings Panel with username and controls

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-87
https://perigongames.atlassian.net/browse/BFB-89
**Description**
1. Created Settings panel
2. Separated Set username panel - forced, and setUsername
3. Placed Control Settings panel into the settings panel

**Screenshots / GIFs**

Screen Shot 2021-11-07 at 8 43 42 AM
Screen Shot 2021-11-07 at 8 42 28 AM

PR #55, opened 2021-11-08
Bfb 41

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-41

**Description**
Added Pause Menu into our scenes

Press Escape to open/close Pause Menu.

Resume,Reset and Quit works as intended! 
Will add Settings Functionality too after  #49 and #54 gets merged into develop!. 

PauseMenu


Assets/_BForBoss/Scripts/Managers/WorldManager.cs
```diff
@@ -74,6 +93,21 @@ private void SetupLeaderboardViews()
             _perigonAnalytics.StartSession(SystemInfo.deviceUniqueIdentifier);
         }
 
+        private void Update()
+        {
+            if (Keyboard.current[PauseKey].wasPressedThisFrame)
```
Gursaanj, line 98: Right now the InputSystem has the Escape key attached to the FirstPersonCharacter, where pressing it auto unlocks the cursor! So kinda makes unpausing a bit wonky here cause the cursor is supposed to lock but sometimes doesn't

If its cool @m4rtint can I remove the InputSystem for Escape?

m4rtint: not yet, it might break some other stuff, i'll handle it when i connect stuff.

m4rtint (approved):
lgtm

m4rtint (approved):
it lgtm

PR #57, opened 2021-11-13
BFB-90: Putting leaderboard into settings

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-90

**Description**
- Placing Leaderboard into settings
- Abstracting Locking Controls when menu is open

**Screenshots / GIFs**
image

### DIFF
--- a/Assets/_BForBoss/Scripts/Managers/WorldManager.cs
+++ b/Assets/_BForBoss/Scripts/Managers/WorldManager.cs
@@ -15,7 +15,6 @@ namespace BForBoss
         [SerializeField] private PauseMenu _pauseMenu;
         [SerializeField] private TimerViewBehaviour _timerView = null;
         [SerializeField] private ForcedSetUsernameViewBehaviour _forcedUploadView = null;
-        [SerializeField] private LeaderboardPanelBehaviour _leaderboardPanel = null;
 
         [Title("Effects")] 
         [SerializeField] private Volume _deathVolume = null;
@@ -28,6 +27,7 @@ namespace BForBoss
         // This probably best placed inside its own utility section
         private readonly StateManager _stateManager = StateManager.Instance;
         private readonly PerigonAnalytics _perigonAnalytics = PerigonAnalytics.Instance;
+        private FreezeActionsUtility _freezeActionsUtility = null;
         // This is probably best kept within its own utility section
         private PostProcessingVolumeWeightTool _postProcessingVolumeWeightTool = null;
         private DetectInput _detectInput = new DetectInput(); //Placeholder, remove this after finishing the timed leader board stuff
@@ -80,14 +80,14 @@ namespace BForBoss
             _player.Initialize();
             _checkpointManager.Initialize(_detectInput, _timeManagerViewModel);
             _timeManager.Initialize(_timeManagerViewModel);
+            _freezeActionsUtility = new FreezeActionsUtility(_player);
         }
 
         private void SetupUserInterface()
         {
-            _pauseMenu.Initialize(_player);
+            _pauseMenu.Initialize(_player, _freezeActionsUtility);
             _timerView.Initialize(_timeManagerViewModel);
-            _leaderboardPanel.Initialize(LockMouseUtility.Instance);
-            _forcedUploadView.Initialize();
+            _forcedUploadView.Initialize(_freezeActionsUtility);
         }
 
         private void OnApplicationQuit()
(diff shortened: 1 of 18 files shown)


Assets/_BForBoss/Scripts/Managers/WorldManager.cs
```diff
@@ -15,7 +15,6 @@ public class WorldManager : MonoBehaviour
         [SerializeField] private PauseMenu _pauseMenu;
         [SerializeField] private TimerViewBehaviour _timerView = null;
         [SerializeField] private ForcedSetUsernameViewBehaviour _forcedUploadView = null;
-        [SerializeField] private LeaderboardPanelBehaviour _leaderboardPanel = null;
```
m4rtint, line 18: Placing Leaderboard into the settings view

Assets/_BForBoss/Scripts/UserInterface/SettingsViewBehaviour.cs
```diff
@@ -7,30 +7,49 @@ namespace BForBoss
...
         [SerializeField] private Button _backButton = null;
+
+        private InputSettingsViewBehaviour _inputSettingsView = null;
+        private SetUsernameViewBehaviour _setUsernameView = null;
+        private LeaderboardPanelBehaviour _leaderboardView = null;
+        private TabbedPanelViewBehaviour _tabbedPanelViews = null;
         
-        private InputSettingsViewBehaviour InputSettingsView => GetComponentInChildren<InputSettingsViewBehaviour>();
-        private SetUsernameViewBehaviour SetUsernameView => GetComponentInChildren<SetUsernameViewBehaviour>();
-        private TabbedPanelViewBehaviour TabbedPanelViews => GetComponentInChildren<TabbedPanelViewBehaviour>();
```
m4rtint, line 13: Certain Views can't be found at times, since the gameobject is not active. So im just setting them on awake

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.InputSettings.cs
```diff
@@ -95,13 +95,21 @@ public float ControllerVerticalSensitivity
 
         public void DisableActions()
         {
+            _dashBehaviour?.DisableAction();
+            crouchInputAction?.Disable();
+            jumpInputAction?.Disable();
+            movementInputAction?.Disable();
```
m4rtint, line 101: All input should be disabled

mikeslgoh (approved):

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.InputSettings.cs
unoctium1: Is it possible to just disable the base `InputActionAsset` here instead? That seems like a slightly more elegant and extendable approach, especially if we need to add new inputs

unoctium1 (commented):
Seems mostly good, but I'm getting two issues: First, hitting escape while the settings or leaderboard are open closes the pause menu and unlocks input, but the settings stays open. Second, I'm not seeing any results on the leaderboard at the moment in editor. I'm not sure if thats an issue with this per se though, or if we cleared the database or something.

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.InputSettings.cs
m4rtint: oh wans't aware of that,i'll try it out.

m4rtint:
@unoctium1 i cleared out the leaderboard

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.InputSettings.cs
m4rtint: Actually. would that also disable the UI movement as well? especially for the controller?

m4rtint:
> Seems mostly good, but I'm getting two issues: First, hitting escape while the settings or leaderboard are open closes the pause menu and unlocks input, but the settings stays open. Second, I'm not seeing any results on the leaderboard at the moment in editor. I'm not sure if thats an issue with this per se though, or if we cleared the database or something.

I semi fixed the Leaderboard when you finish the race. 
But there's a slight low priority bug that not going to fix just yet since it'll probably means adding more states to the statemanager.  Either way, just won't be able to open the pause menu when you finish the race

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.InputSettings.cs
unoctium1: I could be off about this, but I believe the event system takes care of that separately and it should be safe, would be worth testing that first though. We could also rather than disabling the whole asset disable the player controls action map, and then any UI-specific controls would go in their own, new action map

m4rtint: Cleaned it up to disable the Player Control action map

unoctium1 (approved):
Works great, lgtmi

PR #58, opened 2021-11-15
Setup wall running animations

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-71

**Description**
Wall run animations - currently just using the turn/strafe animations already set up, but so that the player is constantly strafing away from the wall. Looks more compelling than it sounds.
Also added some line breaks to wall running 

**Screenshots / GIFs**
https://youtu.be/O8wy2uvXuGM

### DIFF
--- a/Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.Animator.cs
+++ b/Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.Animator.cs
@@ -21,7 +21,11 @@ namespace BForBoss
                 return;
             }
 
-            SetRunSpeed();
+            if (IsWallRunning())
+                SetWallRunSpeed();
+            else
+                SetRunSpeed();
+
 
             animator.SetBool(GroundParamId, IsOnGround() || IsWallRunning());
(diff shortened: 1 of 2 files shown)


Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.Animator.cs
```diff
@@ -61,10 +66,15 @@ private void SetRunSpeed()
...
             float forwardAmount = useRootMotion ? move.z : Mathf.InverseLerp(0.0f, GetMaxSpeed(), GetSpeed());
-            forwardAmount = IsWallRunning() ? 1f : forwardAmount; //always go at max speed when wall running, walk animations look weird on the wall
             
             SetAnimatorSpeed(forwardAmount);
             SetAnimatorTurn(Mathf.Atan2(move.x, move.z));
         }
+
+        private void SetWallRunSpeed()
+        {
+            SetAnimatorSpeed(1f);
```
m4rtint, line 76: can you put this as a const on top and name it apprioriately?

m4rtint (approved):

PR #59, opened 2021-11-15
Fix for Body Visibility Bug

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-93

**Description**
Cull body visibility when in third person, and modify switching so that animation continues during the transition

I also fixed a bug to do with switching views well wall running that was somewhat related.

### DIFF
--- a/Assets/_BForBoss/_Character/FPS_Camera_Blends.asset
+++ b/Assets/_BForBoss/_Character/FPS_Camera_Blends.asset
@@ -35,3 +35,47 @@ MonoBehaviour:
         m_PreInfinity: 0
         m_PostInfinity: 0
         m_RotationOrder: 0
+  - m_From: CM Crouch
+    m_To: CM OverShoulder
+    m_Blend:
+      m_Style: 0
+      m_Time: 0.2
+      m_CustomCurve:
+        serializedVersion: 2
+        m_Curve: []
+        m_PreInfinity: 0
+        m_PostInfinity: 0
+        m_RotationOrder: 0
+  - m_From: CM Walk
+    m_To: CM OverShoulder
+    m_Blend:
+      m_Style: 0
+      m_Time: 2
+      m_CustomCurve:
+        serializedVersion: 2
+        m_Curve: []
+        m_PreInfinity: 0
+        m_PostInfinity: 0
+        m_RotationOrder: 0
+  - m_From: CM OverShoulder
+    m_To: CM Crouch
+    m_Blend:
+      m_Style: 0
+      m_Time: 2
+      m_CustomCurve:
+        serializedVersion: 2
+        m_Curve: []
+        m_PreInfinity: 0
+        m_PostInfinity: 0
+        m_RotationOrder: 0
+  - m_From: CM OverShoulder
+    m_To: CM Walk
+    m_Blend:
+      m_Style: 0
+      m_Time: 2
+      m_CustomCurve:
+        serializedVersion: 2
+        m_Curve: []
+        m_PreInfinity: 0
+        m_PostInfinity: 0
+        m_RotationOrder: 0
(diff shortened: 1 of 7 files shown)


Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -246,16 +264,41 @@ private void ResetJumpCount()
...
         }
 
         private void SwitchView(InputAction.CallbackContext context)
         {
             if (context.started) IsThirdPerson = !IsThirdPerson;
         }
 
+        private IEnumerator TogglePlayerModelWithDelay()
+        {
+            yield return new WaitForSeconds(SwitchViewDuration);
```
m4rtint, line 292: Wondering, why have the need to have a coroutine when toggling? Is it to prevent spamming the button?

m4rtint:
I tested this, and i think the issue is still there, where if u look down while in first person, u still see the players model

m4rtint:
Also thinking of the feasibility of moving the camera stuff into a separate script. - like dash/slide...ect

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -33,6 +42,8 @@ public bool IsThirdPerson
             }
         }
 
+        private float SwitchViewDuration => cmBrain.m_DefaultBlend.m_Time;
```
m4rtint, line 45: Can you also change the blendtime to 0.5?

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
unoctium1: Its so that when you blend from third person to first, you still see the body up until the moment you've completed the blend. In the case of going from first person to third I'm not using the coroutine, I'm just immediately switching

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
unoctium1: Yeah, maybe in that case I should modify this to use a custom blend instead of relying on the default blend

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
unoctium1: This is also why I changed the animator to use a different variable instead of directly relying on the `IsThirdPerson` variable, so that it animates up until the blend is complete

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
unoctium1: I just pushed a change that simplifies the logic to remove the coroutine and smooth transition, and since that means the model isn't visible during the blend I went ahead and replaced it with a cut

unoctium1:
Just drastically simplified this and replaced the transitions with a cut. It looks a little less cool, but makes a lot more sense if we're adding this functionality as a button in the menu instead

Assets/_BForBoss/_Character/FPS_Camera_Blends.asset
```diff
@@ -35,3 +35,47 @@ MonoBehaviour:
...
+        serializedVersion: 2
+        m_Curve: []
+        m_PreInfinity: 0
+        m_PostInfinity: 0
+        m_RotationOrder: 0
+  - m_From: CM OverShoulder
+    m_To: CM Walk
+    m_Blend:
+      m_Style: 0
+      m_Time: 2
```
m4rtint, line 75: is it okay for my to change it from cut to 0.2?

m4rtint (approved):

PR #61, opened 2021-11-19
Player Pref Debug Tool

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-80

**Description**
Create tool to remove player prefs.

Currently allows you to remove all player prefs and individually remove any keys defined in the `UploadPlayerScoreDataSource.PlayerPrefKey` struct


unoctium1:
@m4rtint I know you don't like xml comments but in this case I felt like even though the code is relatively clear the assumptions should probably be made explicit

m4rtint:
> @m4rtint I know you don't like xml comments but in this case I felt like even though the code is relatively clear the assumptions should probably be made explicit

Hmm i mean it's fine, since it's a debug tool, and the annoyance and bad things would fall on us and not the player, so i'm not as stingy. 

However, to alleviate any future annoyances. Change the name of this method to be like `GetStringTypeKeysFromStruct`, + create a ticket to fix it aka, handle non string args (*Optional). You can leave the comment, but 1/10 people will read that comment. That 1 is either you or me.

unoctium1:
Fixed that, and made https://perigongames.atlassian.net/browse/BFB-100

m4rtint (approved):

m4rtint (approved):
Nice, u can probably close the ticket u made.

Assets/_BForBoss/Scripts/Debug/DebugView/PlayerPrefResetter.cs
```diff
@@ -0,0 +1,86 @@
...
+    public class PlayerPrefResetter : DebugView
+    {
+        private Vector2 _scrollPosition = Vector2.zero;
+
+        private IList<string> _keys;
+
+        public PlayerPrefResetter(Rect masterRect) : base(masterRect)
+        {
+            _keys = GetConstStringValuesFromStruct<UploadPlayerScoreDataSource.PlayerPrefKey>();
+            // if we add player pref fields anywhere else make sure to append them to the array here
```
m4rtint, line 18: There's actually more than a few places with PlayerPrefKeys

m4rtint: https://perigongames.atlassian.net/browse/BFB-75

I created a ticket to add them in, for the future

PR #63, opened 2021-11-19
BFB-102: Limit the username

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-102

**Description**
Limits the username to be only alpha numeric.
This regex should now allow alpha number + space

**Screenshots / GIFs**
Capture

### DIFF
--- a/Assets/_BForBoss/_Leaderboard/Scripts/Username/InputUsername.cs
+++ b/Assets/_BForBoss/_Leaderboard/Scripts/Username/InputUsername.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Sirenix.Utilities;
 using UnityEngine;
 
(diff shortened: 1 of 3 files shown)


Assets/_BForBoss/_Leaderboard/Scripts/Username/InputUsername.cs
```diff
@@ -1,3 +1,6 @@
+using System;
+using System.Linq;
```
unoctium1, line 2: I think this is an unneeded using

unoctium1 (changes requested):
Mostly good, but I think its a good idea to clear out the unnecessary linq using before merging this

Assets/_BForBoss/_Leaderboard/Scripts/Username/InputUsername.cs
m4rtint: Fixed

unoctium1 (approved):
lgtm

PR #65, opened 2021-11-22
BFB-75: Restructuring Folders

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-75

**Description**
- Restructuring the folders to include - Utilitiy, UserInterface and Core
- Changed the name space to use `Perigon.wtvr`
- 
**Screenshots / GIFs**
image


Assets/_BForBoss/_Character/Scripts/PlayerSlideBehaviour.cs
```diff
@@ -1,8 +1,7 @@
-using ECM2.Characters;
 using ECM2.Common;
```
unoctium1, line 1: Whats the purpose of removing the character using if we're still using the common namespace?

Assets/_BForBoss/_Core/Scripts/Debug/DebugView/PlayerPrefResetter.cs
```diff
@@ -14,7 +16,9 @@ public class PlayerPrefResetter : DebugView
 
         public PlayerPrefResetter(Rect masterRect) : base(masterRect)
         {
-            _keys = GetConstStringValuesFromStruct<UploadPlayerScoreDataSource.PlayerPrefKey>();
+            _keys = GetConstStringValuesFromStruct<PlayerPrefKeys.InputSettings>();
+            _keys.AddRange(GetConstStringValuesFromStruct<PlayerPrefKeys.ThirdPerson>());
+            _keys.AddRange(GetConstStringValuesFromStruct<PlayerPrefKeys.GameplaySettings>());
```
unoctium1, line 21: I wonder if it would make sense to move the get const sting values function into PlayerPrefKeys, and just have it output a list for the whole class?

Assets/_BForBoss/_Utility/Scripts/PlayerPrefKeys.cs
```diff
@@ -0,0 +1,27 @@
+namespace Perigon.Utility
+{
+    public class PlayerPrefKeys 
+    {
```
unoctium1, line 4: Could this be a static class?

unoctium1 (commented):
This is great! Once we get this merged in I might play around with trying to convert some of the modules into packages, but I'm not sure how feasible that will be. Either way, I think this is a great change

Assets/_BForBoss/_Character/Scripts/PlayerSlideBehaviour.cs
unoctium1: Actually this might get at another point - what if we moved external resources into their respective modules? So ECM2 is part of the _Character module only, instead of just loose in the project. The project I'm on in my day job has a really intense package workflow, and third party assets end up in resource folder inside of their respective packages if they're needed for package functions, it could be something we could consider copying here

Assets/_BForBoss/_Core/Scripts/Debug/DebugView/PlayerPrefResetter.cs
m4rtint: yeah that would be better.

Assets/_BForBoss/_Utility/Scripts/PlayerPrefKeys.cs
m4rtint: I Could change that

Assets/_BForBoss/_Character/Scripts/PlayerSlideBehaviour.cs
m4rtint: so to answer ur first question. 
Because we're naming this `Character` is conflicts with the `ECM2` `Character.

m4rtint: answer to the second question. 
The sole reason to keep it outside, is if we were to update the module, it'll be easier to see what needs to be updated, what doesn't.

Assets/_BForBoss/_Core/Scripts/Debug/DebugView/PlayerPrefResetter.cs
m4rtint: @unoctium1 do you mind checking out this branch and doing that?

unoctium1: Yeah for sure, in that case dw about changing it to static, I can do that as well

Assets/_BForBoss/_Character/Scripts/PlayerSlideBehaviour.cs
unoctium1: Yeah thats fair, I was playing around with it, and well moving ECM2 inside and replacing its asmdef with a reference file worked quite well, I did realise that we'd pretty quickly run into a dependency hell scenario with odin and dotween, since pretty much everything depends on them. We could try to bundle them with utilities since that is also a common dependency, but I think it would pretty quickly be more trouble than its worth

m4rtint: odin and dotween is sort of unavoidable since you do use it everywhere.

unoctium1:
@m4rtint I also added default namespaces to each assembly, new scripts should now be in the right namespace by default

Assets/_BForBoss/_UserInterface/Scripts/UserInterface.asmdef
```diff
@@ -1,6 +1,6 @@
 {
     "name": "UserInterface",
-    "rootNamespace": "",
+    "rootNamespace": "BForBoss",
```
m4rtint, line 3: CAn you rename this to Perigon.UserInterface

unoctium1: Sure, the scripts in that folder aren't using that at the moment though

m4rtint: ah fuck i forgot to rename them. I'll update it

unoctium1 (approved):
lgtm

PR #69, opened 2021-11-29
Object Pooling set up

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-112

**Description**
I decided to model this off of the api for unity 2021's object pooler, so that it is fairly replaceable if we decide to upgrade. For good measure, I pulled it into a 2021 project and made it a wrapper for built in pooling. All pooled objects are kept in their own scene, at root level, for best performance. Creating a new pool requires a name (used for the pool scene), a function to create new objects, and actions to be called upon pulling out/ recycling a new object

**Screenshots / GIFs**
https://youtu.be/C0tRYe4qxz0


Assets/_BForBoss/_Utility/ObjectPoolerSample/SpawnParticles.cs
```diff
@@ -0,0 +1,67 @@
+using Perigon.Utility;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+namespace BForBoss
+{
+    public class SpawnParticles : MonoBehaviour
```
m4rtint, line 7: Is this like a demo script on how to use the `ObjectPooler`?

Assets/_BForBoss/_Utility/ObjectPoolerSample/SpawnParticles.cs
```diff
@@ -0,0 +1,67 @@
+using Perigon.Utility;
+using UnityEngine;
+using UnityEngine.InputSystem;
+
+namespace BForBoss
```
m4rtint, line 5: This should be Perigon.Utiltiy

Assets/_BForBoss/_Utility/ObjectPoolerSample/SpawnParticles.cs
unoctium1: Yeah, I left it in at the moment but I'll remove it before closing this

unoctium1: The whole "ObjectPoolerSample" folder can be removed

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,109 @@
...
+#else
+		protected Scene poolScene;
+		protected List<T> pool = null;
+#endif
+
+		protected Func<T> createFunc;
+		protected Action<T> actionOnGet;
+		protected Action<T> actionOnRelease;
+
+        public ObjectPooler(string poolName, Func<T> createFunc, Action<T> actionOnGet, Action<T> actionOnRelease)
```
m4rtint, line 27: Are these Actions and Functions Nullable?
and what is 
`createFunc` means?

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Collections.Generic;
+#endif
+
+namespace Perigon.Utility
+{
+    public class ObjectPooler<T> where T : Component
+    {
+		protected string name;
+
+#if UNITY_2021_1_OR_NEWER
```
m4rtint, line 16: So Instead of having a bunch of these everywhere, can you have 1 `bool flag` property that get's set as true within this, and then you can do all your code within checks of that flag.

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
m4rtint: Oh okay, so `creatFunc` is like a builder. 
What do you think of the arg name - `objectBuilder`

m4rtint:
So I was wondering: How would clean up work when we e.g. go back to the main menu

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Sure I can make that change, I think an xml comment on the constructor is probably a good idea too

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: I'm not sure that will work, the UnityEngine.Pool api doesn't exist in sub 2021 versions so wrapping it in bools won't work without compiler conditionals

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
m4rtint: As long as Unity doesn't change their ObjectPool constructor, that should be okay.

unoctium1: And yeah good catch, the actions are nullable but the function isn't, I'll make that explicit

unoctium1:
@m4rtint good question, I'll add an explicit destructor to clean up the list. But at the moment it will just get picked up by the garbage collector

m4rtint:
> @m4rtint good question, I'll add an explicit destructor to clean up the list. But at the moment it will just get picked up by the garbage collector

Thanks. 
I trust C# garbage collector, not so much Unity's lol

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
m4rtint: ohh i see. Good to know.

m4rtint: https://perigongames.atlassian.net/browse/BFB-127 

Made a ticket to refactor as reminder if/when we get to 2021

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,109 @@
...
+			if (pool == null)
+			{
+				CreatePools();
+			}
+			actionOnRelease?.Invoke(toRecycle);
+			pool.Add(toRecycle);
+#endif
+		}
+
+#if !UNITY_2021_1_OR_NEWER
```
m4rtint, line 82: why do you need this here if it's going to compile anyway?

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,109 @@
...
+		void CreatePools()
+		{
+			pool = new List<T>();
+
+			if (Application.isEditor)
```
m4rtint, line 87: How come this section is only run in Editor?

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: I mean its not strictly necessary, but just to keep it simpler in Unity 2021 I wrapped all the unneeded scene management stuff in conditional flags. This code should never run in 2021, so I figure it makes sense to just keep it out

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Its specifically to support domain reload in the editor. If the domain reload triggers, the list gets lost, so this just handles repopulating it. Come to think of it I should wrap this in editor compiler flags

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,130 @@
...
+			objectPool.Dispose();
+#else
+			GameObject[] rootObjects = poolScene.GetRootGameObjects();
+			for (int i = 0; i < rootObjects.Length; i++)
+			{
+                UnityEngine.Object.Destroy(rootObjects[i]);
+			}
+			pool = null;
+#endif
+		}
```
unoctium1, line 58: I'm not sure how necessary this is cause I think it gets done anyways when the scene unloaded, but I'm not really sure

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
m4rtint: Does it still need `if (Application .isEditor)`?

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -0,0 +1,126 @@
...
+			this.buildObject = buildObject;
+			this.actionOnGet = actionOnGet;
+			this.actionOnRelease = actionOnRelease;
+
+#if UNITY_2021_1_OR_NEWER
+			objectPool = new ObjectPool<T>(buildObject, actionOnGet, actionOnRelease);
+#endif
+		}
+
+        ~ObjectPooler()
```
m4rtint, line 46: didn't know C# has a deinit.

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Good catch I'll remove that

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Yeah haha they're pretty rarely used

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
m4rtint: That's fine for now, just can be a bit confusing readability wise.

m4rtint (approved):
lgtm

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Yeah tabs and color formatting are doing a lot of work, I'm not sure if you're just reading it on github but it is a bit better off of github

PR #71, opened 2021-12-03
BFB-120: Crosshair

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-120

**Description**
*Note* - There are more than a few ways to implement crosshair, after some googling, i'm going with the most simple one since I'm still experimenting with what works, what doesn't. 
Call this iteration 1
- Added few crosshairs
- Added script to set crosshair
- created prefab for crosshair

**Screenshots / GIFs**
image


Assets/_BForBoss/_Weapons/Scripts/CrosshairBehaviour.cs
```diff
@@ -0,0 +1,28 @@
...
+        private void Start()
+        {
+            _crosshair.sprite = _defaultCrosshair;
+        }
+
+        [Button]
+        public void SetCrosshairImage(Sprite image)
+        {
+            _crosshair.sprite = image;
+        }
```
unoctium1, line 26: Is the point of this just so we can quickly test different sprites?

unoctium1 (approved):
lgtm, I'm not sure I get the point of CrosshairBehaviour but thats not blocking at all

Assets/_BForBoss/_Weapons/Scripts/CrosshairBehaviour.cs
m4rtint: yup

m4rtint:
> lgtm, I'm not sure I get the point of CrosshairBehaviour but thats not blocking at all

Originally, had the idea of having the crosshair dynamically enlarging and shrinking when you hold down the trigger.  And some other stuff, but that's a bit overcomplicating things, so just sticking with iteration 1 - simple first.

unoctium1:
> Originally, had the idea of having the crosshair dynamically enlarging and shrinking when you hold down the trigger. And some other stuff, but that's a bit overcomplicating things, so just sticking with iteration 1 - simple first.

Sounds good, I think having the script in place will definitely make things better if we want to do anything more complicated later

PR #74, opened 2021-12-07
BFB-113: Bullet prefabs and spawning

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-113

**Description**
Create `BulletBehaviour`, `BulletProperties` scriptable, and bullet spawning logic (using object pooling). At the moment, there are two different bullet prefabs, `RifleBullet_Physics` and `RifleBullet_NoPhysics`. Physics uses rigidbodies + colliders, and NoPhysics does manual collision detection. Both have their upsides and downsides, so I've left both in the project at the moment. I also added `BulletTrail`, which controls a trail renderer on the bullet, mostly for debug purposes at the moment


m4rtint:
Oh this is good, keep both. cuz one could be a "grenade launcher" type deal

Assets/_BForBoss/_Weapons/Scripts/RifleBulletBehaviour_NoPhysics.cs
```diff
@@ -0,0 +1,40 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Perigon.Weapons
+{
+    public class RifleBulletBehaviour_NoPhysics : BulletBehaviour
```
m4rtint, line 8: No need to call the specifically Rifle. most bullets behavior are the same
But name is `NoPhysicsBulletBehaviour`

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        RifleNoPhysics,
+        RiflePhysics
+    }
+    
+    public class BulletSpawner : MonoBehaviour
+    {
+        [SerializeField] private BulletBehaviour[] _bulletPrefabs;
+        [InlineEditor]
+        [SerializeField] private BulletPropertiesScriptableObject _bulletPropertiesScriptableObject;
```
m4rtint, line 18: Can you set it up so that the Bullet itself have the scriptable object?

Assets/_BForBoss/_Weapons/Prefabs/RifleBullet_NoPhysics.prefab
```diff
@@ -0,0 +1,79 @@
+%YAML 1.1
```
m4rtint, line 1: Look through the prefabs, wondering why there's a cylinder and a sphere mesh on this.

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
```diff
@@ -0,0 +1,52 @@
...
+using Sirenix.OdinInspector;
+using UnityEngine;
+
+namespace Perigon.Weapons
+{
+
+    public enum BulletTypes
+    {
+        RifleNoPhysics,
+        RiflePhysics
```
m4rtint, line 11: No need to name these specifically Rifle.

m4rtint: Do we need an enum for the Bullets type? since it's really only, either Yes Physics, or No Physics.

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
m4rtint: Actually rethinking about this. 
WeaponBehaviour -> right now includes a bullet spawner. 

Ideally, the weaponScriptableObject would include the bullet it wants to spawn and shoot out.
So i would think the weapon would pass in the desired bullet into the Bullet spawner

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+    
+    public class BulletSpawner : MonoBehaviour
+    {
+        [SerializeField] private BulletBehaviour[] _bulletPrefabs;
+        [InlineEditor]
+        [SerializeField] private BulletPropertiesScriptableObject _bulletPropertiesScriptableObject;
+        // placeholder for more robust bullet switching
+        [SerializeField] private bool _usePhysics = false;
```
m4rtint, line 20: I think to simplify things, we can remove this. 

And then the designer can just pass in the correct `BulletBehaviour` Prefab of their choice.  Instead of having an array of `BulletBehaviour`

The Physics bullet behaviour is currently out of scope. But it's a nice to keep in mind for the future.

Assets/_BForBoss/_Weapons/Prefabs/RifleBullet_NoPhysics.prefab
unoctium1: image
I figured this looked like a decent placeholder bullet

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
unoctium1: Sure, so move it out of initialize and then just have it set itself up on awake? Should I just get rid of the interface in that case?

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
unoctium1: I was kind of thinking we could change `BulletSpawner.SpawnBullet` to take a `BulletType` parameter, and then give a `BulletType` property to the `WeaponScriptableObjects` - do you that would work, or is that overengineering it a bit?

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
unoctium1: Not necessarily, no - this was more future proofing for rockets/grenades/etc

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
unoctium1: Thinking about this further, I'm leaning towards your initial suggestion of having the bullet prefab hold its own scriptable and then having `weaponScriptableObject` hold a bullet type, since that would make for the most robust & self contained object pools.

I could also make it so `weaponScriptableObject` holds the bullet scriptable, and then passing that in every time a new bullet is spawned, but I think we'd start to lose any benefits to that if we decide a specific weapon needs a specific bullet prefab

m4rtint: So your idea is to have 
1. Separate enum for bulletType on the weapon
2. Whatevre bulletType is passed into bulletSpawner, the BulletSpawner creates the prefab which contains the bulletScriptable.

The issue is that, whenever I want to create a new bullet to iterate/test/develop, i would need to update the enum. Which can cause issue since you'll need to keep track of the Enum as well as the BulletPrefab + Scriptable

unoctium1: Well you wouldn't necessarily run into too many issues there, internally the enum is just an index in the array of prefabs. If you're iterating you just need to make sure you have the same number of enum options and prefabs, and I can make a custom editor to enforce that

unoctium1: I'm just trying to avoid coupling bullets to weapons too tightly since it feels redundant to have multiple object pools for identical weapons

unoctium1: If you'd prefer, the enum was really only for readability. We could just use an int directly and then the issue of needing to update the enum goes away

m4rtint: alright that makes sense. 
Would it be more visually easier to map, if you either have a serialized Dictionary<enum, Prefab> 
or a serializable struct with the enum and prefab.

unoctium1: Went ahead and made a custom inspector, this also ensures the array is always the correct length and to the end user doesn't show as an array

m4rtint: Okay if this is the case, can you make NoPhysicsBullet its own prefab instead of a variation of the original
image

m4rtint: Just to make sure we're on the same page, Depending on the designs of bullets/guns. we would have 
Shotgun bullets
rocket bullets
Automatic bullets
Single shot rifle bullets
pistol bullets
sniper bullets.

 So the enum would potentially include these in the future.

unoctium1: I'd say yes, if those bullets are fundamentally different (use different bullet behaviour scripts or a different prefab). If we can reuse the same bullet for multiple weapons we should wherever possible

m4rtint: I'm thinking for mostly the visual appearance of the bullets could be different

unoctium1: Yeah in that case they'll be different types

unoctium1: @m4rtint just submitted a fix for the prefab, both bullets are their own root prefabs now

Assets/_BForBoss/_Weapons/Scripts/BulletSpawner.cs
```diff
@@ -0,0 +1,45 @@
+using Perigon.Utility;
+using UnityEngine;
+
+namespace Perigon.Weapons
+{
+
+    public enum BulletTypes
```
m4rtint, line 7: Can you move this to a separate file.

m4rtint: And then this PR should be good to go

m4rtint (approved):

PR #77, opened 2021-12-08
BFB-95: Test cases for wall running

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-95

**Description**
- Added Test Scene for wall running
- Added tests for wall running
- Not every case is added, but the base is there

**Screenshots / GIFs**
image


unoctium1 (approved):
This is great, lgtm

PR #80, opened 2021-12-13
BFB-117 - Death Fade Out

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-117

**Description**
Add dissolve fade out integrated with death animation

**Screenshots / GIFs**
https://youtu.be/s5hsNINSdEw

### DIFF
--- a/Assets/_BForBoss/_Entities/Scripts/DummyTargetBehaviour.cs
+++ b/Assets/_BForBoss/_Entities/Scripts/DummyTargetBehaviour.cs
@@ -5,7 +5,7 @@ namespace Perigon.Entities
 {
     public class DummyTargetBehaviour : LifeCycleBehaviour
     {
-        private const float DEATH_ANIMATION_DURATION = 5f;
+        
         private const string IS_DEAD = "IsDead";
         private const string HIT = "Hit";
         private const float MAX_DISSOLVE = 1f;
@@ -13,7 +13,8 @@ namespace Perigon.Entities
         private readonly int HIT_ID = Animator.StringToHash(HIT);
         private readonly int DISSOLVE_ID = Shader.PropertyToID("_Dissolve");
         
-        
+        [SerializeField]
+        private float _dissolveVFXDuration = 7.5f;
         
         private Animator _animator;
         private Renderer _renderer;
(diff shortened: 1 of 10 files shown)


Assets/_BForBoss/_Entities/Scripts/DummyTargetBehaviour.cs
```diff
@@ -9,22 +8,29 @@ public class DummyTargetBehaviour : LifeCycleBehaviour
...
             base.Awake();
             _animator = GetComponentInChildren<Animator>();
+            _renderer = GetComponentInChildren<Renderer>();
         }
 
         protected override void LifeCycleFinished()
         {
             _animator.SetBool(DEATH_ID, true);
-            Destroy(gameObject, DEATH_ANIMATION_DURATION);
+            var tween = _renderer.material.DOFloat(MAX_DISSOLVE, DISSOLVE_ID, DEATH_ANIMATION_DURATION * 2);
```
m4rtint, line 31: Why is there a * 2 here?

m4rtint:
Can you change the health property to be 2, instead of 50. So it's easier to test.

Assets/_BForBoss/_Entities/Scripts/DummyTargetBehaviour.cs
unoctium1: Setting it to the same time as the animation, I found it faded out a little too fast and you couldn't really see the whole animation

unoctium1: I could expose that, probably a better idea than hardcoding it

m4rtint: you can name it `DISSOLVE_ANIMATION_DURATION`,  and just set it to the time of your liking.  We can think about the Tween duration separate from the death anim duration.

m4rtint: Like it's good to time it with the death, but the value you're using will still belong to the Dissolve tween

m4rtint (approved):

PR #81, opened 2021-12-13
BFB-139: Weapon Swap Mechanic

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-139

**Description**
- EquipmentBehaviour added
- Weapon Swap with scroll wheel added
- Controller North Button to swap weapons added

Left Out:
Using Keys to swap weapons, since it's not needed until we design how many weapon slots we need.

**Screenshots / GIFs**


Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
```diff
@@ -0,0 +1,160 @@
...
+                Debug.LogWarning("There are currently no weapons equipped within Equipment Behaviour");
+            }
+            
+            SetupPlayerEquipmentInput();
+        }
+        
+        #region Input 
+        private void OnMouseSwapWeaponAction()
+        {
+            var scrollVector = _isMouseScrollEnabled ? Mouse.current.scroll.ReadValue().normalized : Vector2.zero;
```
m4rtint, line 140: @unoctium1 this is the only big problem i have with this task/new input system.
Mouse.current.scroll.ReadValue(), i can't for the life of my get the scroll wheel working.

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            if (_inputActions == null)
+            {
+                Debug.LogWarning("Input Action Asset is missing from equipment behaviour");
+            }
+
+            if (_weapons.IsNullOrEmpty())
+            {
+                Debug.LogWarning("There are currently no weapons equipped within Equipment Behaviour");
+            }
```
Gursaanj, line 132: More of question, how do we feel about using the [OnValidate()](https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnValidate.html) Unity Message for null checking components?  Just an idea

So that checking every awake, we ensure the devs make sure no component is null

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
```diff
@@ -0,0 +1,160 @@
...
+        private void SetupPlayerEquipmentInput()
+        {
+            _reloadInputAction = _inputActions.FindAction("Reload");
+            _fireInputAction = _inputActions.FindAction("Fire");
+            _swapWeaponInputAction = _inputActions.FindAction("WeaponSwap");
+        }
+
+        private void SetupControllerSwapWeaponInputBinding()
+        {
+            _swapWeaponInputAction.started += OnControllerSwapWeaponAction;
```
Gursaanj, line 85: Does this ever need to unsubscribe?

Gursaanj (approved):

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
m4rtint: yeah we could, either would work. Especially for value changes on inspector
Might be cleaner overall to separate from `Awake()`
https://docs.unity3d.com/ScriptReference/MonoBehaviour.OnValidate.html

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
m4rtint: technically no.  But i'll set it up OnDestroy()

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
unoctium1: I'm not sure I understand, what you have seems to be working for me? It might be an issue for adding controller bindings but this works

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            if (_fireInputAction != null)
+            {
+                _fireInputAction.Disable();
+            }
+            
+            if (_swapWeaponInputAction != null)
+            {
+                _swapWeaponInputAction.Disable();
+            }
```
unoctium1, line 61: Real minor thing but could we replace the enable/disable setup with null propagation? Just seems unnecessarily large as it is

Assets/_BForBoss/_Weapons/Scripts/EquipmentBehaviour.cs
m4rtint: So instead of adding the "mouse scroll" within the New Input System, I'm called the Mouse directly. So I am not able to set it up through the Input Action Asset

unoctium1: Did you see this thread? https://forum.unity.com/threads/how-do-you-get-mouse-scroll-input.825672/

m4rtint: yup, looked that up and tried their settings.

unoctium1: We could just call this good enough for now and then make a big ticket? It sounds like this might be an issue on Unity's end

PR #84, opened 2021-12-16
BFB-135: Reloading UI Feedback

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-135

**Description**
- Reloading UI updates the reloading duration

**Screenshots / GIFs**
demo1


unoctium1 (commented):
Code looks good, but there are some strange issues on the automatic rifle specifically. Reloading only seems to be working when holding down fire, and if you start reloading by hitting the reload key while bullets are still available, it only gets part of the way through reloading before cancelling the reload and shooting again

Edit: Looking at the code for the AutomaticWeapon, it looks like reloading is behind `if(!_isFiring) return` and its not for semi auto weapons

m4rtint:
> Code looks good, but there are some strange issues on the automatic rifle specifically. Reloading only seems to be working when holding down fire, and if you start reloading by hitting the reload key while bullets are still available, it only gets part of the way through reloading before cancelling the reload and shooting again
> 
> Edit: Looking at the code for the AutomaticWeapon, it looks like reloading is behind `if(!_isFiring) return` and its not for semi auto weapons

ahhhh i see, i'll fix

Assets/_BForBoss/_Weapons/Scripts/AutomaticWeaponBehaviour.cs
```diff
@@ -22,14 +22,13 @@ protected override void OnFireInputAction(InputAction.CallbackContext context)
 
         protected override void Update()
         {
-            if (!_isFiring)
+            if (_isFiring)
             {
-                return;
+                _weapon.DecrementElapsedTimeRateOfFire(Time.deltaTime);
```
unoctium1, line 27: Shouldn't this also decrement even if not firing?

m4rtint: yup you are correct, updating

unoctium1 (approved):
Great, lgtm

Assets/_BForBoss/_Core/Scripts/ReloadViewBehaviour.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        private void Update()
+        {
+            _reloadView.fillAmount = GetElapsedDurationPercentageDone();
+        }
+
+        private float GetElapsedDurationPercentageDone()
+        {
+            if (_equipmentData != null)
```
Gursaanj, line 30: do you want to add a check to make sure that `_equipmentData.MaxReloadDuration` cannot be 0?

Gursaanj (approved):
Looks good, small comment

PR #85, opened 2021-12-17
BFB-141: Rotate towards wall direction on wall run

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-141

**Description**
Now pushes the player towards the wall direction when wall running, also some small code cleanup on wall running

### DIFF
--- a/Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
+++ b/Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
@@ -14,6 +14,8 @@ namespace Perigon.Character
         private float _maxWallRunAcceleration = 20f;
         [SerializeField]
         private float _wallGravityDownForce = 0f;
+        [SerializeField] 
+        private bool _factorLookDirectionIntoVelocity = false;
 
         [Header("Wall Run Conditions")]
         [SerializeField, Tooltip("Stop a wall run if speed dips below this")]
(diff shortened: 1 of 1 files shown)


m4rtint:
It feels very good this time around.

m4rtint:
How difficult is it to keep the max wall run speed even when looking closer to the wall?

unoctium1:
Do you just mean ensure the velocity is independent of where the camera is looking?

m4rtint:
> Do you just mean ensure the velocity is independent of where the camera is looking?

Yeah, more of a curiosity, not sure if it'll be better or worse.

unoctium1:
> > Do you just mean ensure the velocity is independent of where the camera is looking?
> 
> Yeah, more of a curiosity, not sure if it'll be better or worse.

It should be pretty easy, I'll try making it toggleable and see how well it works

unoctium1:
Tried it out, its maybe harder than I'd thought. Seems like its much harder to stay attached to the wall and I'm not really sure why

m4rtint:
> Tried it out, its maybe harder than I'd thought. Seems like its much harder to stay attached to the wall and I'm not really sure why

yeah i think it's better to remove it. 

I wonder  if there's a way to use a curve for the Look along the wall motion, so even if it's super fast look speed, it won't be too jerky.

unoctium1:
> I wonder if there's a way to use a curve for the Look along the wall motion, so even if it's super fast look speed, it won't be too jerky.

What do you mean? Like a curve for the rotation?

m4rtint:
> > I wonder if there's a way to use a curve for the Look along the wall motion, so even if it's super fast look speed, it won't be too jerky.
> 
> What do you mean? Like a curve for the rotation? 

So currently when rotating towards forward,  it's pretty much a linear curve, I'm wondering if it's possible to do another type of curve.

unoctium1:
> So currently when rotating towards forward, it's pretty much a linear curve, I'm wondering if it's possible to do another type of curve.

I'm not sure how feasible that would be, its currently just getting the difference between the heading and current and adding the rotation each frame. Adding a curve would be a pretty drastic restructuring, and I'm not sure it would still respond to mouse input as well

m4rtint (approved):

PR #86, opened 2021-12-17
BFB-143: Refactor Input Settings Menu

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-143

**Description**
- Refactor the Input settings to split between controller and mkb

**Screenshots / GIFs**
image


Assets/_BForBoss/_Core/Scripts/UserInterface/SettingsViewBehaviour.cs
```diff
@@ -57,7 +62,20 @@ private void Awake()
             _backButton.onClick.AddListener(ClosePanel);
             SetupViews();
         }
-        
+
+        private void OnValidate()
```
Gursaanj, line 66: As an aside, if we are cool with moving these kinda calls to `OnValidate()` then we should create a tech debt (backlog) task to refactor current classes to do the same. For consistency

Gursaanj (approved):
lgtm

Assets/_BForBoss/_Core/Scripts/UserInterface/SettingsViewBehaviour.cs
m4rtint: There aren't that many on awake currently? i don't think? so we can change them when we see them. Starting  now.

PR #89, opened 2021-12-31
Gursaanj/resolver tool

Gursaanj:
**Description**
At the current moment each field with the `[Resolve] `attribute will have an altered property drawer with 4 buttons (Resolve via children, via parent, via self, and the clear current field).

It currently works for normal object reference fields, fields as a part of arrays/lists as well as fields within structs (or as arrays/Lists defined in structs) but I am pretty sure it will break with more complex nesting of fields (i.e arrays of structs that have arrays) but I will worry about that once we finalize if we like this tool/find it useful.

Also for a strange bug related to property drawers (I can explain in detail next time we speak to each other) Don't put the `[Resolve]`  Attribute on a field for a struct. Rather, put in on the fields declared within the struct (as can be seen in ResolveTestScript.cs ).
When you have the time play around in the `ResolveTestScene `scene

**Screenshots / GIFs**
image

image


m4rtint:
The [Resolve] Currently doesn't work outside of `_utility` because it's outside the asmdef file. 
you need to either put the PropertyAttribute folder into the Scripts folder, or create a separate asmdef. 

I also would like to see how Odin does their asmdef

m4rtint:
In terms of inactive components, this is how Odin sets up their attributes. 
idk if it's nice to follow something similiar for inactive components. 
Imo, it would be nice to default to Including Inactive

Screen Shot 2021-12-31 at 12 47 50 PM

m4rtint:
Disable it if it's a scriptableObject?

we can leave it to iteration 2

Screen Shot 2021-12-31 at 12 57 55 PM

m4rtint:
I'm trying to use Resolver in structs, just like how you did it, but it's not working. 

I read that. you don't want to allow Resolver on structs? Is that the long term goal? 

Screen Shot 2021-12-31 at 1 03 16 PM
Screen Shot 2021-12-31 at 1 03 40 PM

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+        private bool _isContentInitialized = false;
+        private bool _includeInactiveGameObjects;
+        
+        private bool _isList = false;
+        private bool _isArray = false;
+        private bool _isDeclaredInStruct = false;
+        
+        //Button Content
+        private GUIContent _childResolveContent;
+        // private GUIContent _sceneResolveContent;
```
m4rtint, line 32: Are all these needed or...

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+            EditorGUI.PropertyField(position,property,label);
+
+            CreateRects(position);
+
+            DrawResolverButtons(property);
+        }
+
+        private void DrawResolverButtons(SerializedProperty property)
+        {
+            // Self (G) | Children (C) | Parent (P) | (OPTIONAL) Sibilings (R) | None (X)
```
m4rtint, line 96: Can you update this comment. 
Comments are dangerous cuz if doesn't change. 👻

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+                EditorGUI.PropertyField(position, property, label);
+                return;
+            }
+
+            if (!_isContentInitialized)
+            {
+                InitializeContent();
+
+                ResolveAttribute resolveAttributeAttribute = (ResolveAttribute) attribute;
+                _includeInactiveGameObjects = resolveAttributeAttribute.IncludeInactive;
```
m4rtint, line 69: How do you include Inactive currently? are we able to change it?

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+            bool isGameObjectType = fieldType == typeof(GameObject);
+
+            List<GameObject> gameObjects = new List<GameObject>();
+            Component[] components = null;
+
+            switch (resolveType)
+            {
+                case ResolveType.FromChildren:
+                {
+                    if (isGameObjectType)
```
m4rtint, line 136: In terms of cleaning up, you can always abstract this out into separate methods so it's more readable

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+                            .Where(component => component.gameObject != mb.gameObject).ToArray();
+                    }
+                    
+                    break;
+                }
+                case ResolveType.FromParent:
+                {
+                    if (isGameObjectType)
+                    {
+                        Transform parent = mb.transform.parent;
```
m4rtint, line 155: NiT: maybe rename this, since it's confusing. like `mbParent`

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+                }
+            }
+        }
+
+        private void InitializeContent()
+        {
+            _childResolveContent = new GUIContent("C", "Get Component from Children");
+            _parentResolveContent = new GUIContent("P", "Get Component from Parent");
+            _selfResolveContent = new GUIContent("S", "Get Component from Self");
+            _clearContent = EditorGUIUtility.IconContent("d_winbtn_mac_close_h@2x", "|Clear Component");
```
m4rtint, line 238: >"|Clear Component"
Do you need the "|"

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+                    property.serializedObject.Update();
+                    property.objectReferenceValue = components[0];
+                    property.serializedObject.ApplyModifiedProperties();
+                }
+            }
+        }
+
+        private void InitializeContent()
+        {
+            _childResolveContent = new GUIContent("C", "Get Component from Children");
```
m4rtint, line 235: "Get Component(s) from Children"

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -0,0 +1,369 @@
...
+
+            bool isSerializeField = fieldInfo.GetCustomAttribute(typeof(SerializeField)) != null;
+            return (fieldInfo.IsPublic || ((fieldInfo.IsFamily || fieldInfo.IsPrivate) && isSerializeField));
+        }
+
+        private bool IsAppropriateType(SerializedProperty property)
+        {
+            Type fieldType = fieldInfo.FieldType;
+            
+            //Todo: Find a long term solution to not allow attribute being placed on structs moving forward
```
m4rtint, line 281: Can you make a ticket for this as a reminder. and link the ticket here

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
unoctium1: If I'm reading it right there's an optional constructor and it defaults to true, so if you don't want to include inactive children you'd go `[Resolve(false)]`

unoctium1:
@m4rtint my understanding is that your mistake with the struct is putting the [Resolve] attribute on the struct itself, you'd need it on the fields within the struct only. Imo this makes sense, since structs can't ever be complete scene references themselves

m4rtint:
> @m4rtint my understanding is that your mistake with the struct is putting the [Resolve] attribute on the struct itself, you'd need it on the fields within the struct only. Imo this makes sense, since structs can't ever be complete scene references themselves

But in the image I posted, I only set the resolve on an int..
 Which I realize doesn't make sense. But shouldn't really show the error.

unoctium1:
@m4rtint its the `[Resolve][SerializeField] private SomethingElse _something` line that's giving the error. The `int` field and the struct itself is set up correctly I believe

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Yup exactly, it will default to search inactive gameobjects. and you can toggle that off with the optional parameter

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Yeah in reality, both `Components` and `GameObjects` derive from `Object` so I shouldn't really need two separate implementations. And instead of using `GetInstanceID()` to get the ID for ObjectReference I could use `GetHashCode` which Object class implements.

Could possibly fix the `ScriptableObject` derivation issue as well now that I think about it

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Yeah. To be honest I have no idea why haha!

https://docs.unity3d.com/ScriptReference/EditorGUIUtility.IconContent.html

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Well you are only resolving to one component, so shouldn't we use the singular in this case?

Gursaanj:
> @m4rtint its the `[Resolve][SerializeField] private SomethingElse _something` line that's giving the error. The `int` field and the struct itself is set up correctly I believe

Yup exactly!! the `[Resolve]` attribute can't be put on field declared structs (i.e. `_something`) but you can put it on the individual fields declared within the struct itself (i.e. `SliderBehaviour`)

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Probably not, was going to make a button for Sibling gameObjects but not sure if the use case is used enough to justify it

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: > Disable it if it's a scriptableObject?
> 
> we can leave it to iteration 2
> 
> Screen Shot 2021-12-31 at 12 57 55 PM

Hmm good point, will look into it in a couple of hours, and will let you know if we should leave it to iteration 2. Could be the same case as structs, where the fields themselves can have the attribute but the actual ScriptableObject declared field cannot

m4rtint:
> > @m4rtint its the `[Resolve][SerializeField] private SomethingElse _something` line that's giving the error. The `int` field and the struct itself is set up correctly I believe
> 
> Yup exactly!! the `[Resolve]` attribute can't be put on field declared structs (i.e. `_something`) but you can put it on the individual fields declared within the struct itself (i.e. `SliderBehaviour`)

Ahh I see, that makes sense. 
I'm wondering instead of giving an error, u can just ignore it.

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: oh in my headd, you are showing all the components in the children. but what you wrote is fine too.

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: >character to flag it as a tooltip.
what the fuck, why not just build it in.

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: How difficult is it to implement it like the Odin style? 
`[Resolve(IncludeInactive = false)]`

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Actually, @m4rtint  How is it supposed to work for ScriptableObjects? Since they are not Components or Gameobjects they cant just exist as a child, parent or self if that makes sense, They will always be a field of a component, instead of the actual component

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Wait im still confused, what is the difference between how it is now ,
image,

And how you want it to be?

m4rtint: I didn't see it. Ita fine as it is then

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: Nothing should happen when a [Resolve] is on the scriptableobjects

Assets/_BForBoss/_Utility/Scripts/PropertyAttributes/ResolveAttribute.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using UnityEngine;
+
+namespace BForBoss
```
m4rtint, line 4: Rename this to 
Perigon.Utility

m4rtint:
Everything works great outside even after export - only issue is the name space. 
QoL for next time:
1. Would be nice if you can get this sort of error printed out for [Resolve] Struct issue, instead of printing in the log continuously 
Screen Shot 2022-01-03 at 1 30 42 AM

2. Another QoL suggestion- would be cool to enable an attribute to use `FindObjectOfType<>` 
So if you have a root manager object that needs some buttons inside another Root Canvas for example. Be able to find that. 

I'll approve when namespace is fixed.

Assets/_BForBoss/_Utility/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: https://perigongames.atlassian.net/browse/BFB-153

Assets/_BForBoss/_Utility/Scripts/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -153,21 +153,24 @@ private ComponentInformation[] GetResolvedGameObjects(MonoBehaviour mb, ResolveT
             {
                 case ResolveType.FromChildren:
                 {
-                    foreach (Transform child in mb.transform)
+                    Transform[] transforms = mb.GetComponentsInChildren<Transform>(_includeInactiveGameObjects)
```
m4rtint, line 156: What are these changes for?

Assets/_BForBoss/_Utility/Scripts/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
```diff
@@ -153,21 +153,24 @@ private ComponentInformation[] GetResolvedGameObjects(MonoBehaviour mb, ResolveT
...
-                        components.Add(new ComponentInformation(child.gameObject));
+                        components.Add(new ComponentInformation(transform.gameObject));
                     }
                     
                     break;
                 }
                 case ResolveType.FromParent:
                 {
-                    Transform mbParent = mb.transform.parent;
+                    Transform[] transforms = mb.GetComponentsInParent<Transform>(_includeInactiveGameObjects)
```
m4rtint, line 168: If I understand correctly, it finds the component within the Siblings of the parent?
A
  - B
   - - C [Resolve]
  - D
  - E
  - F

So it tries to find B, D, E F?

Assets/_BForBoss/_Utility/Scripts/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
m4rtint: A
    - B
     --C [Resolve]
       --- D
          ----H
       --- E
          ----F
          ----G


So does it try to find all of D,H,E,F,G? 
Where as before it goes for only D, H?

Gursaanj: Basically I realised that the way I had before did not account for children of children (which is how GetComponentsInChildren works) so it was inconsistent with the functionality when trying to resolve components instead of GameObjects. 

So in this case it would find D,H,E,F,G (like how the component one would work). This way also allows the `IncludeInactive` parameter to take into account for the gameObject search as well

Assets/_BForBoss/_Utility/Scripts/Editor/PropertyDrawers/ResolverPropertyDrawer.cs
Gursaanj: Nope no siblings only the direct trail of parents (I.e. B and A, assuming B is a child of A)

m4rtint (approved):

m4rtint (changes requested):
The build fails currently

unoctium1:
@Gursaanj just took a look at the branch, pretty sure the issue is that the Editor folder in Utility is missing its own asmdef file. Creating an editor only file there should fix builds

m4rtint:
> @Gursaanj just took a look at the branch, pretty sure the issue is that the Editor folder in Utility is missing its own asmdef file. Creating an editor only file there should fix builds

I tried that, i don't think it'll work, since it still needs reference to unity.IMGUI, which i can't find the reference to.
A possible work around would be to wrap everything into 
`#if UNITY_EDITOR`

unoctium1:
> I tried that, i don't think it'll work, since it still needs reference to unity.IMGUI, which i can't find the reference to. A possible work around would be to wrap everything into `#if UNITY_EDITOR`

I'm not sure I get what you mean, how would an edtor only asmdef be missing imgui? I tried creating an editor asmdef and didn't run into any issues, but I don't think I really understand the problem you're describing either

m4rtint:
> > I tried that, i don't think it'll work, since it still needs reference to unity.IMGUI, which i can't find the reference to. A possible work around would be to wrap everything into `#if UNITY_EDITOR`
> 

> I'm not sure I get what you mean, how would an edtor only asmdef be missing imgui? I tried creating an editor asmdef and didn't run into any issues, but I don't think I really understand the problem you're describing either

Oh i miss understood what you meant. 
Screen Shot 2022-01-04 at 10 41 29 AM
But this is what meant right/

unoctium1:
@m4rtint 

Yeah that's all I meant, it built when I tested locally on my end

m4rtint (approved):
Lgtm

PR #93, opened 2022-01-08
BFB-153 Custom Validator Added

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-153

**Description**
Created Inspector based message to indicate the field-declared structs with the `[Resolve]` attribute can't be visualized and should be amended.

**Screenshots / GIFs**
image

### DIFF
--- a/Assets/_BForBoss/_Utility/Scripts/Editor/AttributeValidators/ResolveAttributeValidator.cs
+++ b/Assets/_BForBoss/_Utility/Scripts/Editor/AttributeValidators/ResolveAttributeValidator.cs
@@ -1,0 +1,24 @@
+using System;
+using Sirenix.OdinInspector.Editor.Validation;
+
+[assembly: RegisterValidator(typeof(Perigon.Utility.ResolveAttributeValidator))]
+
+namespace Perigon.Utility
+{
+    public class ResolveAttributeValidator : AttributeValidator<ResolveAttribute>
+    {
+        protected override void Validate(ValidationResult result)
+        {
+            Type fieldType = Property.Info.TypeOfValue;
+
+            //Check if field type is Struct
+            if (fieldType.IsValueType && !fieldType.IsPrimitive && fieldType != typeof(decimal) &&
+                fieldType != typeof(DateTime) && !fieldType.IsEnum)
+            {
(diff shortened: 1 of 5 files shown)


Assets/_BForBoss/_Utility/Scripts/Editor/AttributeValidators/ResolveAttributeValidator.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Perigon.Utility
+{
+    public class ResolveAttributeValidator : AttributeValidator<ResolveAttribute>
+    {
+        protected override void Validate(ValidationResult result)
+        {
+            Type fieldType = Property.Info.TypeOfValue;
+
+            //Check if field type is Struct
```
m4rtint, line 14: NiT:  if you don't wanna write this comment, you can put whatever is in the if statement into a separate bool 
```suggestion
           var isFieldTypeStruct = fieldType.IsValueType && !fieldType.IsPrimitive && fieldType != typeof(decimal) &&
                fieldType != typeof(DateTime) && !fieldType.IsEnum
```

m4rtint:
Please put 
BFB-153 into your title

Assets/_BForBoss/_Utility/Scripts/Editor/AttributeValidators/ResolveAttributeValidator.cs
Gursaanj: Fair sounds good, I was actually thinking of making an extension method for `Type` class and name it `IsStruct`, instead of this mess. What do you think @m4rtint

m4rtint: My at work staff dev sort of made us adopt a rule of 3.

If you found that it's used in 3 different places, then extract it out.

m4rtint: Just a tip, but it's ultimately up to you.

m4rtint (approved):

PR #95, opened 2022-01-14
BFB-124: Weapon Bob and Sway

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-124

**Description**
- Rearranged Some files within the Weapon Folder
- Added a Weapon Camera to only render the weapon
- Added new Layer for Layer mask for weapon
- Added Bob and Sway to weapon

Main File to Check with all the new logic:
Scripts/Weapons/WeaponsManager.cs 

**Screenshots / GIFs**
https://www.youtube.com/watch?v=mhSSm8blzUM


Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponsManager.cs
```diff
@@ -0,0 +1,131 @@
...
+        private bool CanBobWeapon()
+        {
+            return (_characterMovement.IsWallRunning || _characterMovement.IsGrounded) 
+                   && !_characterMovement.IsDashing 
+                   && !_characterMovement.IsSliding;
+        }
+
+        private void UpdateWeaponRecoil()
+        {
+            if (_weaponRecoilLocalPosition.z >= _accumulatedRecoil.z * 0.99f)
```
unoctium1, line 118: Probably should pull this out to a const, maybe the numbers in vBobValue too though those are less of a big deal imo

unoctium1 (commented):
This looks great, the bob feels pretty good and the recoil makes all the weapons feel a lot better. I do kinda wonder about the bobbing while wall running, it feels strange that its the same as regular walking. Probably more something to consider for a second iteration than part of this PR though

m4rtint:
> 

yeah it was really hard to pin point what was wrong exactly, but it maybe if we separate the speed depending on the state, then imight feel better

Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponsManager.cs
m4rtint: yeah i don't actually know what these values are for, i got them from unity's tuts. Il'l pull them out and try to name it appropriately

Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponsManager.cs
```diff
@@ -0,0 +1,131 @@
...
+            {
+                characterMovementFactor =
+                    Mathf.Clamp01(characterVelocity.magnitude / _characterMovement.CharacterMaxSpeed);
+            }
+            
+            _weaponBobFactor = Mathf.Lerp(_weaponBobFactor, characterMovementFactor, _weaponBobSharpness * Time.deltaTime);
+            
+            var hBobValue = Mathf.Sin(Time.time * _weaponBobFrequency) * _hipFireBobAmount * _weaponBobFactor;
+            /// Trignometric Graph Tranformation: y = A * Sin(b * x - c) + d
+            var vBobValue = (Mathf.Sin(Time.time * _weaponBobFrequency * 2f) * 0.5f + 0.5f) * _weaponBobFactor * _hipFireBobAmount;
```
m4rtint, line 102: @unoctium1 only part with hardcoded values, since i honestly am not sure how to name them. 
But I'M GUESSING it's for Trig graph transforms

unoctium1: Yeah it looks like its just constraining the sine values to be between 0 and 1. I'm fine leaving these ones, it was really just the 0.99 that bothered me

unoctium1:
> yeah it was really hard to pin point what was wrong exactly, but it maybe if we separate the speed depending on the state, then imight feel better

Yeah I kinda vote we leave this for a second iteration and maybe just make a ticket for now, I think we'll probably need to brainstorm what effects we want while wall running first

unoctium1 (approved):
lgtm

PR #99, opened 2022-01-24
BFB-110: FMOD Integration

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-110

**Description**
Add the fmod unity plugin to the project, and add a sample fmod project as BeForBoss/FMODAssets


m4rtint:
I see that you placed the FMODAssets in the Root folder, any reasons why it's in the Root folder instead of the `Asset` folder?

m4rtint:
have u ever seen this error? idk if i should ignore or fix, but i ignored for now

unoctium1:
> I see that you placed the FMODAssets in the Root folder, any reasons why it's in the Root folder instead of the `Asset` folder?

Yeah, that's what I was talking about on Slack - that's an actual FMOD project, and we explicitly don't want Unity to import it. However, using a path inside the project lets the Unity plugin use a relative project path which hopefully should prevent errors

unoctium1:
>  have u ever seen this error? idk if i should ignore or fix, but i ignored for now

No I haven't seen that, are you working on a mac? I think if its just line endings repairing them shouldn't be an issue, but I'd worry we'll keep doing that back and forth

m4rtint:
> > I see that you placed the FMODAssets in the Root folder, any reasons why it's in the Root folder instead of the `Asset` folder?
> 
> Yeah, that's what I was talking about on Slack - that's an actual FMOD project, and we explicitly don't want Unity to import it. However, using a path inside the project lets the Unity plugin use a relative project path which hopefully should prevent errors



> >  have u ever seen this error? idk if i should ignore or fix, but i ignored for now
> 
> No I haven't seen that, are you working on a mac? I think if its just line endings repairing them shouldn't be an issue, but I'd worry we'll keep doing that back and forth

it's on Windows, if it keeps happening we can look up a solution

m4rtint:
Just wanted to play around with it a bit more before merging, But this is how you interact with the new FMOD project? 
image

unoctium1:
> Just wanted to play around with it a bit more before merging, But this is how you interact with the new FMOD project?

Yeah, its how you preview it within unity. To actually edit you'd need to open the *.fspro file in fmod, but for our purposes using the event browser is probably all you'd need to do. Currently its a super simple placeholder project, which is why there's just one event. I'll probably make it more complex when I add samples

m4rtint (approved):

PR #101, opened 2022-01-26
BFB-173: Create FMOD samples

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-173

**Description**
Create a scene demonstrating fmod samples. Includes a script showing how to play audio and modify parameters, as well as a audio source that plays on start and an example of timeline integration. Also includes an "Advanced Samples" button that directs users to [Fmod's scripting examples](https://www.fmod.com/resources/documentation-unity?version=2.02&page=examples.html)

**Screenshots / GIFs**
image


Assets/FMOD Samples/Scripts/FMODBasicSamples.cs
```diff
@@ -0,0 +1,45 @@
...
+            //RuntimeManager.PlayOneShot(); 
+            // This is the FMOD equivalent of giving a script an audio clip and then playing it using an audio manager instead of a direct audio source,
+            // but it really complicates parameters
+        }
+        
+        public void PlayWithPitch()
+        {
+            var pitch = _pitchSlider.value;
+            _emitter.Play(); // note - play clears cached parameter values. Call this, THEN set params!
+            _emitter.SetParameter(PARAM_NAME, pitch);
```
m4rtint, line 37: so is this parameter set for the _emitter above?

unoctium1: It has a default value of 1 set in the inspector but otherwise its just getting set here

m4rtint (changes requested):
The build currently breaks, so i'm going to wait until that's resolved before merging this in

Assets/FMOD Samples/Scripts/FMODBasicSamples.cs
```diff
@@ -0,0 +1,45 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Perigon.FMODExamples
+{
+    public class FMODBasicSamples : MonoBehaviour
+    {
+        private const string PARAM_NAME = "Pitch";
+        
+        [SerializeField] Slider _pitchSlider;
```
Gursaanj, line 12: should have `private` access modifier

m4rtint (approved):

PR #103, opened 2022-01-28
BFB-148: Bullet hitmarker

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-148

**Description**
- Added hitmarker png
- Added event Actions for the bullets. 
- I really hope I mem managed it correctly.

**Screenshots / GIFs**
https://www.youtube.com/watch?v=X4ENJszxst4


Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletBehaviour.ApplyDamage.cs
```diff
@@ -10,6 +10,8 @@ protected void HitObject(Collider col, Vector3 hitPosition, Vector3 hitNormal)
             if(col.TryGetComponent(out LifeCycleBehaviour lifeCycle))
             {
                 lifeCycle.Damage(_properties.Damage);
+                var isDead = !lifeCycle.IsAlive;
```
Gursaanj, line 13: If it is for readability and all that's fair, but not sure if it's worth caching a local variable if it's only being used once?

Assets/_BForBoss/_Weapons/Scripts/Weapons/CrosshairBehaviour.cs
```diff
@@ -7,23 +8,74 @@ namespace Perigon.Weapons
...
+        {
+            _elapsedHitMarkerTime -= Time.deltaTime;
+            _elapsedKillMarkerTime -= Time.deltaTime;
+            if (_elapsedKillMarkerTime > 0)
+            {
+                _hitMarker.color = Color.red;
+            } 
+            else if (_elapsedHitMarkerTime > 0)
+            {
+                _hitMarker.color = Color.white;
```
Gursaanj, line 50: Do you want to make these colors a serialized field variable, so we can change it from the inspector later down the line if need be

Assets/_BForBoss/_Weapons/Scripts/Weapons/CrosshairBehaviour.cs
```diff
@@ -7,23 +8,74 @@ namespace Perigon.Weapons
...
+            }
+            else
+            {
+                _hitMarker.color = Color.clear;
+            }
+        }
+        
+        private void Awake()
```
Gursaanj, line 58: Instead of Awake() should this be set as an Initialize method, that gets called from `WeaponBehaviour`, since that class already touches this `OnEnable`. It might be better for Script execution order

m4rtint: I think my line of Thinking is that, if this class is not initialize from outside, it'll still show the default image.

So it's like a backup, and some sort of crosshair still shows.
In terms of script execution order, we know that initialize is always called on start, since that's what we agreed upon. So set default will always be called first.

Assets/_BForBoss/_Weapons/Scripts/Weapons/CrosshairBehaviour.cs
m4rtint: It might not even be color in the future, these are almost placeholder. I could extract these out to be read-only variables?

Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletBehaviour.ApplyDamage.cs
m4rtint: Yeah sure

Assets/_BForBoss/_Weapons/Scripts/Weapons/CrosshairBehaviour.cs
Gursaanj: If its a placeholder, then I leave it up to you! Yeah maybe a Read-Only would be better?

Gursaanj (approved):
lgtm

PR #104, opened 2022-01-28
BFB-134 : Sandbox SceneTemplate

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-134

**Description**
Created a scene template in which users can create a new scene with all the bare bone assets/prefabs needed to play/build a scene with a working Character and Pause menu Functionality.

Refer to the gif below as to how to create a new Scene from this template (will be put onto Jira ticket as well). 

Note that this dialog does not exist from creating a scene based on the project inspector (i.e. right-click -> New Scene). So for now, the only way is to click on File -> New Scene

**Screenshots / GIFs**
Create from New Scene Template


m4rtint:
You left a message refering to leaving out DebugView, If you think that's a good judgement call and you tested it out and it works better that way, then that's okay 👍

m4rtint (approved):
Everything looks good. 

Maybe ping @unoctium1  just incase,

unoctium1 (changes requested):
This looks mostly good to me, could you replace the AudioListener on the main camera with an FMOD Event listener though? I can also make that change quickly if you're prefer

Gursaanj:
> This looks mostly good to me, could you replace the AudioListener on the main camera with an FMOD Event listener though? I can also make that change quickly if you're prefer

Good Catch, just updated the template scene

unoctium1 (approved):
lgtm

PR #105, opened 2022-01-29
BFB-181 : Use Explicit Interface implementation

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-181

**Description**
The implementation for the following interfaces has been changed to use the Explicit Implementation as by our new coding standards. Note that each individual commit from 57cf8b5 and onwards focuses on a single interface implementation. Though I have tested all of them I would like if someone could test to ensure that `IThirdPerson`, `ILeaderboardGetEndPoint`, `ILeaderboardPostEndPoint` have been implemented correctly and functionality works as intended.

`ILifeCycle` implementation was a bit awkward and was wondering if I could place `LifeCycle.HealBy`, `LifeCycle.DamageBy` and `LifeCycle.Reset` into the ILifeCycle implementation? @m4rtint. Seems weird that the functionality exists `LifeCycle` and not `ILifeCycle`

### List of Interfaces

- IWeaponProperties
- ICharacterSpeed
- ICharacterSpawn
- IThirdPerson
- IStateManager
- ISpecification
- ILifeCycle
- IHealth
- ILeaderboardGetEndPoint
- ILeaderboardPostEndPoint
- IInputSettings
- ILockInput
- IBullet
- IEquipmentData
- IBulletProperties
- ICharacterMovement


m4rtint:
Did you research on the actual use cases of explicit vs implicit interface implementations?  As opposed to just coding standard for readability?  Like the pros and cons..ect 

I'm looking through the code and all the Up casting is very strange.  There's a time and place for explicit interfaces, and from what I can read online, it shouldn't be a basis for the coding standard.

m4rtint:
I was actually under the impression, that it was literally a naming convention, and that logic didn't have to change. 
From reading online, there are a few people out there that only use explicit interfaces from the get go to push decoupling to its max, but what i've seen here is that we're changing it for the wrong reasons.  It's sort of like, setting everything public just cuz it's more "standardized". 

There's no wrong or right answer to this, since it's just how you design it, but I think the biggest issue right now, is that we didn't design our code base with "explicit interface" only mindset, which might cause issues down the road. 

But what I'm certain now, is that the way Microsoft teaches you explicit usage - is when you have 2+ interfaces on the same concrete class, and they both have the same method/property names. 

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/interfaces/explicit-interface-implementation
```
public interface IControl
{
    void Paint();
}
public interface ISurface
{
    void Paint();
}
```

I'm going to keep reading online about this, but this is definitely a great learning opportunity.
@unoctium1 @Gursaanj @mikeslgoh

Gursaanj:
> 

For sure, looking at it, this change seems to enforce decoupling so that if a user is calling a class, they need to be extremely specific if they want it for the functionality that is based on the class or one of the interfaces that it inherits from.

You are right in that this is more of a structural change and that if we would actually like to adopt this then, we would have to refactor a lot of logic and how it calls functionality. So I am down if we, in the end, scrap this coding standard change for the sake of convenience.

PR #108, opened 2022-02-07
Gursaanj/BFB-195 : SceneSwticher EditorWindow

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-191

**Description**
Made a SceneSwitcher Editor Window that will allow users to more quickly go between scenes or load them additively on top of current scenes. Only scenes under the `_BForBoss` subfolder are added into this editor window.

**Screenshots / GIFs**
SceneSwitcher

All scenes under the same parent folder are grouped together. 
The Scene that is highlighted green indicates the current main scene the user is on at the moment.

SceneSwitcherContextMenu

Anytime a user creates, deletes, or renames a scene, the Editor Window, if open, should update automatically. If for some reason, changes have not been updated on the Editor Window,  a manual `Refresh Scenes` button can be used in the context menu.

Favoriting Scenes

A feature I added, when I originally thought we wanted to have all scenes in the Assets folder included for the Editor Window (i.e. the 20+ ECM2 scenes and such) I created a **Favoriting** Feature where user can set their favourite scenes at the top of the window so they don't have to backtrack each time they want to change the scenes they will be working on the most.
***Note** The favourited scenes are saved via `EditorPrefs` so each individual can have their own favourite scenes without influencing any other users.

I definitely see this feature being way more useful in the coming months when our scene count is much higher.


m4rtint:
just need to fix the title of the PR - @mikeslgoh wondering if you had checks in place to make sure the title is correct.

Also not sure if they files should be within the BForBoss Utilities folder instead of the Root Asset folder

Assets/SceneSwitcher/Source/Editor/SceneSwithcerEditorWindow.cs
```diff
@@ -0,0 +1,262 @@
...
+using UnityEngine;
+using UnityEditor;
+using System.IO;
+using PerigonGames;
+using UnityEditor.SceneManagement;
+using UnityEngine.SceneManagement;
+
+public class SceneSwithcerEditorWindow : EditorWindow, IHasCustomMenu
+{
+    private const string PROEJCT_PARENT_FOLDER = "BForBoss/";
```
unoctium1, line 12: Just a small typo here

Assets/SceneSwitcher/Source/Editor/SceneSwithcerEditorWindow.cs
```diff
@@ -0,0 +1,262 @@
...
+    private void OpenSceneSingleMode(SceneConfigSetup sceneConfig)
+    {
+        EditorSceneManager.OpenScene(sceneConfig.path, OpenSceneMode.Single);
+        _currentSceneConfig = sceneConfig;
+    }
+
+    private void LoadSymbols()
+    {
+        _notFavouriteSymbol = EditorGUIUtility.IconContent("d_Favorite On Icon").image as Texture2D;
+        _favouriteSymbol = EditorGUIUtility.IconContent("d_Favorite Icon").image as Texture2D;
```
unoctium1, line 248: Should the strings here be pulled out as consts?

Assets/SceneSwitcher/Source/Editor/SceneSwithcerEditorWindow.cs
unoctium1: And on the class name too actually

unoctium1 (commented):
This looks great, I really like the favoriting toggle feature

Assets/SceneSwitcher/Source/Editor/SceneSwithcerEditorWindow.cs
Gursaanj: haha good eye!

m4rtint (approved):

PR #110, opened 2022-02-09
BFB-184: Fix for object pooler and scene switching

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-188

**Description**
Should fix the object pooler scene switching bug, fps should be consistent after switching the scene now

### DIFF
--- a/Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
+++ b/Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
@@ -37,11 +37,11 @@ namespace Perigon.Utility
 			this.buildObject = buildObject;
 			this.actionOnGet = actionOnGet;
 			this.actionOnRelease = actionOnRelease;
-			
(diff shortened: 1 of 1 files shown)


Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -37,19 +37,22 @@ public ObjectPooler(string poolName, Func<T> buildObject, Action<T> actionOnGet
 			this.buildObject = buildObject;
 			this.actionOnGet = actionOnGet;
 			this.actionOnRelease = actionOnRelease;
+			
+			SceneManager.activeSceneChanged += OnActiveSceneChanged;
```
m4rtint, line 41: Does this also have to be wrapped in a $if statement?

unoctium1: Yeah it does, good catch

m4rtint (approved):

PR #118, opened 2022-02-17
BFB-210: Setup analytics to track which world your playing in

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-210

**Description**
- Use gameObject.scene.name to pass which scene the gameobject belongs to

**Screenshots / GIFs**
Screen Shot 2022-02-21 at 7 42 55 PM


Assets/_BForBoss/_Analytics/Scripts/BForBossAnalytics.cs
```diff
@@ -114,5 +123,20 @@ public void EndSession()
...
+            {
+                case WorldNameAnalyticsName.MS1_RaceCourse:
+                    return EventConstant.MS1_RaceCourse;
+                case WorldNameAnalyticsName.MS2_RaceCourse:
+                    return EventConstant.MS2_RaceCourse;
+                case WorldNameAnalyticsName.Weapons_RaceCourse:
+                    return EventConstant.Weapons_RaceCourse;
+                default:
+                    return EventConstant.Unknown;
+            }
```
unoctium1, line 139: What about just calling toString on the enum? It seems a bit redundant to me to have to have both constant strings and an enum

m4rtint: It's a formatting thing with snake case and lowercase

unoctium1: What about just `ToString().ToLower()` then?

m4rtint: Technically speaking I didn't write the enum correctly in terms of naming standard, which is why we normally map it.

It would be ms1RaceCourse

m4rtint: But if someone else were to write a new course, they would know to write it in this snake case format. Since it's independent from The enum

unoctium1: I see, that makes sense then

m4rtint: I fixed the enums and the const variable naming

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -20,6 +24,7 @@ public abstract class BaseWorldManager : MonoBehaviour
 
         protected abstract Vector3 SpawnLocation { get; }
         protected abstract Quaternion SpawnLookDirection { get; }
+        protected abstract void SetupAnalytics();
```
Gursaanj, line 27: I'm thinking that it might be cleaner to make this a virtual method in which it implements the `DeathAreaBehavior` Initialization and just overrides the method to do nothing for the test scenes (`sandbox`, `GunRange `and etc). Since it looks like this is the primary logic that all proper scenes will most likely have.

```suggestion
        protected virtual void SetupAnalytics()
        {
            FindObjectsOfType<DeathAreaBehaviour>().ForEach(area => area.Initialize(_worldNameAnalytics));
        }
```

Assets/_BForBoss/_Analytics/Scripts/BForBossAnalytics.cs
```diff
@@ -32,22 +35,28 @@ namespace Perigon.Analytics
...
     }
+    
+    public enum WorldNameAnalyticsName
+    {
+        MS1RaceCourse,
+        MS2RaceCourse,
+        WeaponsRaceCourse,
+        Unknown
+    }
     #endregion
```
Gursaanj, line 46: I don't think this is a sustainable or scalable solution, to be honest. Cause now everytime a new scenes is made or renamed, this Enum has to be updated as well. Also making the switch statement way too long in the future.

A potential idea is to just to get the scene name for when the name is needed. 
i.e. `SceneManager.GetActiveScene().name` (something like that) and use that instead. And go ahead and change all current scene names to match the convention you think works best. And Ensure that naming standard is used going forward with any new scenes we will create

@unoctium1  what do you think about this as well?

m4rtint: I changed it to use `gameObject.scene.name` instead. It's much more isolated and cleaner that way.

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
m4rtint: I basically removed the setup analytics completely. just better to let each component handle their own analytics if it's needed

Assets/_BForBoss/_Analytics/Scripts/BForBossAnalytics.cs
```diff
@@ -18,7 +18,10 @@ namespace Perigon.Analytics
     }
     public readonly struct EventConstant
     {
-        public const String RaceCourse = "racecourse";
+        public const String MS1RaceCourse = "ms1_racecourse";
+        public const String MS2RaceCourse = "ms2_racecourse";
+        public const string WeaponsRaceCourse = "weapons_racecourse";
+        public const String Unknown = "unknown";
```
Gursaanj, line 24: These can probably be removed then

Assets/_BForBoss/_Analytics/Scripts/BForBossAnalytics.cs
Gursaanj: Sounds good

Assets/_BForBoss/_Analytics/Scripts/BForBossAnalytics.cs
m4rtint: Removed

Gursaanj (approved):
Looks good

PR #119, opened 2022-02-19
BFB-157: Hide POV one prod build

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-157

**Description**
- Added new DropDownBeheviour as a wrapper for TMP DropDown and Text.
- Hid POV selector on prod

**Screenshots / GIFs**

Screen Shot 2022-02-19 at 1 05 10 PM


unoctium1 (changes requested):
Currently giving a null reference exception in MS1_RaceCourse, WeaponsRaceCourse, and I didn't check but I would guess also the gun range and movement sandbox. I kinda wonder if we should just remove the drop down all together at this point, as without adjusting the gun display third person is already pretty broken

m4rtint:
> Currently giving a null reference exception in MS1_RaceCourse, WeaponsRaceCourse, and I didn't check but I would guess also the gun range and movement sandbox. I kinda wonder if we should just remove the drop down all together at this point, as without adjusting the gun display third person is already pretty broken

I basically cleaned out third person stuff from the GamePlaySettings View 
made a new Ticket for Cleaning out third person completely. And adding free roam camera insted.

https://perigongames.atlassian.net/browse/BFB-211

unoctium1 (commented):
This code looks good to me and yeah all the errors are gone now. That said, I think this isn't solving the actual bug now, the drop down is still there in prod builds

unoctium1 (approved):
lgtm

PR #120, opened 2022-02-21
BFB-201: Add LifeCycleManager and integrate with scenes

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-201

**Description**
Fix for enemy resetting bug in GunRange. Pulls entity specific code into separate manager. 

I do kinda wonder if this needs to be a monobehaviour, I'm open to input on how best to structure this.


Assets/_BForBoss/_Entities/Scripts/LifeCycleManager.cs
```diff
@@ -0,0 +1,26 @@
...
+        public int LivingEntities => _lifeCycleBehaviours.Count(life => life.IsAlive);
+
+        public void Initialize()
+        {
+            _lifeCycleBehaviours = FindObjectsOfType<LifeCycleBehaviour>();
+        }
+
+        public void Reset()
+        {
+            if (_lifeCycleBehaviours == null) return;
```
Gursaanj, line 19: `return` should probably be wrapped within curly braces

Gursaanj (approved):
Looks good

Assets/_BForBoss/_Entities/Scripts/LifeCycleManager.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class LifeCycleManager : MonoBehaviour
+    {
+        private LifeCycleBehaviour[] _lifeCycleBehaviours = null;
+
+        public int LivingEntities => _lifeCycleBehaviours.Count(life => life.IsAlive);
+
+        public void Initialize()
+        {
+            _lifeCycleBehaviours = FindObjectsOfType<LifeCycleBehaviour>();
```
m4rtint, line 14: you can probably put this into Awake instead since `FindObjectsOfType` is a Unity dependent action. So you don't have to way for the Chain of `Initialize` from the top to call this.

PR #124, opened 2022-02-24
BFB-192: Movement Audio

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-192

**Description**
Add state machine for handling walking audio, and events for sliding. dashing, and jumping. Note that you can't hear anything if the background audio is playing. To fix this select the BackgroundAudio object underneath the main camera in any scene and disable it

**Screenshots / GIFs**
https://youtu.be/2jAws2DIO1I


m4rtint:
Can you move the Background Audio prefab to the Root of the hierachy? i find it misleading it to put it under the camera

Assets/_BForBoss/_Character/Scripts/CharacterMovementAudio.cs
```diff
@@ -0,0 +1,161 @@
...
+            if (!_player.IsWallRunning())
+            {
+                var newState = MovementSoundState.Running;
+                PlayState(newState);
+                return newState;
+            }
+            return MovementSoundState.WallRunning;
+        }
+
+        private void PlayState(MovementSoundState state)
```
m4rtint, line 135: can you rename this to something like `PlayStateAudio`

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
```diff
@@ -21,6 +22,9 @@ public partial class FirstPersonPlayer : FirstPersonCharacter
         private PlayerSlowMotionBehaviour _slowMotionBehaviour = null;
 
         [SerializeField] private bool _isThirdPerson = false;
+        
+        public event Action Dashed;
+        public event Action Slid;
```
m4rtint, line 27: can u rename these as `OnDash`
`OnSlide`

Assets/_BForBoss/_Character/Scripts/PlayerSlideBehaviour.cs
```diff
@@ -16,6 +17,7 @@ public class PlayerSlideBehaviour : MonoBehaviour
         private ECM2.Characters.Character _baseCharacter = null;
 
         public bool IsSliding => _isSliding;
+        public event Action StartSliding;
```
m4rtint, line 20: Place this above the properties

Assets/_BForBoss/_Character/Scripts/CharacterMovementAudio.cs
```diff
@@ -0,0 +1,161 @@
...
+        {
+            RuntimeManager.PlayOneShot(_slideAudio, transform.position);
+        }
+        
+        private void PlayDashSound()
+        {
+            RuntimeManager.PlayOneShot(_dashAudio, transform.position);
+        }
+
+        private MovementSoundState CheckNotPlaying()
```
m4rtint, line 80: Can you abstract these `Check___` method, there seems to be copied code within methods. 

Also the method name is misleading as to what is does.  
`Check___` I was assuming it would return some Bool/or void, Where as it's playing audio as well as returning a State.  So the method is doing two in one (playing audio is the side effect)

Assets/_BForBoss/_Character/Scripts/FirstPersonPlayer.cs
unoctium1: So with these I was matching the other actions in ecm2, their conventions are entirely just past tense forms, eg. `Jumped`, `Crouched`, etc

unoctium1:
> Can you move the Background Audio prefab to the Root of the hierachy? I find it misleading it to put it under the camera

Not really, there might be something we can change in the fmod event settings but at the moment because of the spatial aspect it will fall off if the camera moves away from it. Easiest solution imo is to just make it a child of the camera

m4rtint:
> > Can you move the Background Audio prefab to the Root of the hierachy? I find it misleading it to put it under the camera
> 
> Not really, there might be something we can change in the fmod event settings but at the moment because of the spatial aspect it will fall off if the camera moves away from it. Easiest solution imo is to just make it a child of the camera

Okay that makes senes, then no need to touch it. since it's already child of the camera

Assets/_BForBoss/_Character/Scripts/CharacterMovementAudio.cs
unoctium1: I pulled out the side effect so now it should be a bit more clear what these are doing, and renamed the methods as such

m4rtint (approved):

PR #125, opened 2022-02-26
BFB-196: No Debug Window in Prod Build

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-196

**Description**
Debug window will now be destroyed if in a scene during Production Build. When we are sure if a WorldManager will be in every scene (which is not currently the case) I will potentially make that handle conditional Instantiation of the prefab.

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
@@ -34,6 +34,10 @@ namespace BForBoss
         
         private void Awake()
         {
+#if (!UNITY_EDITOR && !DEVELOPMENT_BUILD)
+            Destroy(gameObject);
+#endif
+            
             if (_instance != null && _instance != this)
             {
                 Destroy(gameObject);


m4rtint:
PLease set the PR name correctly

m4rtint:
>When we are sure if a WorldManager will be in every scene (which is not currently the case) I will potentially make that handle conditional Instantiation of the prefab.

Which scenes are missing a world manager? 
And i don't think we need a world manager in every scene, it should be possible that some components are ran themselves.

 my thought process is that, each component shouldn't be 100% coupled to a world manager.

Gursaanj:
> > When we are sure if a WorldManager will be in every scene (which is not currently the case) I will potentially make that handle conditional Instantiation of the prefab.
> 
> Which scenes are missing a world manager? And i don't think we need a world manager in every scene, it should be possible that some components are ran themselves.
> 
> my thought process is that, each component shouldn't be 100% coupled to a world manager.

Right now it is just Leaderboard scene, but like in the future Im sure the title screen will probably not have one.

I'm fine with the lack of coupling, but it feels weird that the DebugWindow should just exist by itself rather than being initialized (depending on if the build is a development or a production build. But we can talk about that in future milestones I guess.

m4rtint:
> > > When we are sure if a WorldManager will be in every scene (which is not currently the case) I will potentially make that handle conditional Instantiation of the prefab.
> > 
> > 
> > Which scenes are missing a world manager? And i don't think we need a world manager in every scene, it should be possible that some components are ran themselves.
> > my thought process is that, each component shouldn't be 100% coupled to a world manager.
> 
> Right now it is just Leaderboard scene, but like in the future Im sure the title screen will probably not have one.
> 
> I'm fine with the lack of coupling, but it feels weird that the DebugWindow should just exist by itself rather than being initialized (depending on if the build is a development or a production build. But we can talk about that in future milestones I guess.

I see this tool similiar to the tools you download from the Unity Store. Where you just drag and drop it into a scene and it just works.

m4rtint (approved):

PR #126, opened 2022-02-28
BFB-205 : Decouple SceneSwitcher from BForBoss with Settings Provider

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-205

**Description**
Created a Settings provider (under Project Settings) that will let users change the Root Folder that the Scene Switcher will use to query for scenes. With this, there is no hardcoded instance of  "BForBoss" in the SceneSwitcher and will allow the tool to be used in different projects without any issue.

Ps. When in project settings you can search for the Settings provider with the keywords ("Scene", "Switch", "Change")

**Screenshots / GIFs**
SettingsProvider


ProjectSettings/SceneSwitcherSettings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "baseSearchFolder": "C:/Users/Gursaanj/Documents/GitHub/BForBoss/Assets/_BForBoss"
+}
```
m4rtint, line 3: Can you put this file into .gitignore

m4rtint:
image

I don't know how much more work it'll be, 

But if the file path is not found, can you render the `choose directory` button on the scene switcher? 
Or a workaround, link a button to open up the Scene Switcher settings to show the user where to change the directory?

m4rtint:
> image
> 
> I don't know how much more work it'll be,
> 
> But if the file path is not found, can you render the `choose directory` button on the scene switcher? Or a workaround, link a button to open up the Scene Switcher settings to show the user where to change the directory?

Maybe also a refresh button, cuz i can make changes, but it won't refresh until I close the Scene switcher panel, and reopen it.

m4rtint:
Works great on mac

Gursaanj:
> > image
> > I don't know how much more work it'll be,
> > But if the file path is not found, can you render the `choose directory` button on the scene switcher? Or a workaround, link a button to open up the Scene Switcher settings to show the user where to change the directory?
> 
> Maybe also a refresh button, cuz i can make changes, but it won't refresh until I close the Scene switcher panel, and reopen it.

image

So the refresh button already exists in the Context Menu. But I just added a "Change Settings"  button in the context menu as well so that it can direct you to the specific project settings page for you to change the directory.

I also just added the change In which if a user were to change the root folder in the settings page, the editor window will automatically refresh (if the editor window is open)

ProjectSettings/SceneSwitcherSettings.json
Gursaanj: But then if we were to change the root folder, how would we make sure that change is pushed for the rest of the team? Don't think it should be a problem if we leave it as is

m4rtint: So as of right now, if this changes, it affects everyone else, 
So it doesn't matter if the root folder changes. I'll still be getting `"C:/Users/Gursaanj/Documents` as my path.

unoctium1: Wouldn't it make more sense to build this dynamically? ie. store /Assets/_BForBoss in the json and then concatenate it with the project folder? Alternatively if we're going to store the whole path it would probably be better suited for the EditorPrefs

Gursaanj: Yeah My bad the path should'nt have that.

Actually, I think this should be better for the UserSettings (instead of ProjectSettings) and maybe place that file in the gitignore?

I was thinking of using EditorPref but if we end up having more settings (not sure what exactly) that scaling that with EditorPrefs for each setting instead of just using a json might be a bit much.  

Since the Json is being Lazily initialized anyways it should be fine!

Gursaanj: Ok nvm Looks like the preferences Folder sets things by `EditorPreferences ` Will just use that

Gursaanj:
Moved Settings Provider from `Project Settings` to `User Preferences`, since the root folder choice should be based on individual preference. The root folder is being saved via `EditorPrefs`

m4rtint (approved):

PR #134, opened 2022-03-05
BFB-223 : Added Entity Kill Counter for WeaponRaceCourse World

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-223

**Description**
Added an active kill counter in the weapon racecourse scene. Where Killing an entity will increment the counter by 1. resetting the scene or restarting it after completion will reset the counter.

**Screenshots / GIFs**
image


m4rtint:
I thought it was supposed to be counting down. 

like 10/10 count down to 0/10 
just to let the players know how many there are.

Assets/_BForBoss/_Entities/Scripts/LifeCycleBehaviour.cs
```diff
@@ -17,6 +18,11 @@ public abstract class LifeCycleBehaviour : MonoBehaviour
 
         public bool IsAlive => _lifeCycle.IsAlive;
 
+        public void AddLifeCycleSubscriber(LifeCycleManager manager)
+        {
+            _lifeCycle.OnDeath += manager.LivingEntitiesAmountChanged;
```
m4rtint, line 23: I think it's better to put this subscription within the `LifeCycle` class instead.

m4rtint: And also food for thought. 

Should each Living Entity need to know about the existance of a manger? or should it be the other way around?

unoctium1: Yeah I think currently the lifecycle doesn't know about the manager and I think we should aim to keep it that way, in my opinion

Assets/_BForBoss/_Core/Scripts/UserInterface/EntityCounterViewBehaviour.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _entitiesDestroyed = 0;
+            _entityCounterLabel.text = _entitiesDestroyed.ToString();
+        }
+
+        private void HandleLivingEntitiesAmountChanged()
+        {
+            _entitiesDestroyed++;
+            _entityCounterLabel.text = _entitiesDestroyed.ToString();
+        }
```
unoctium1, line 27: It might make more sense to keep track of the number killed within the manager, and just access it here, just in case we end up wanting to use it elsewhere. I think that way you could also remove the coupling between the manager and individual lifecycles that way, too, for Martin's point

unoctium1:
> I thought it was supposed to be counting down.
> 
> like 10/10 count down to 0/10 just to let the players know how many there are.

I think in the meeting we discussed that and landed on doing it the way Gursaanj has it, so that you explicitly wouldn't let player's know how many there were

Assets/_BForBoss/_Entities/Scripts/LifeCycleBehaviour.cs
Gursaanj: Fair, will make the change.

Assets/_BForBoss/_Core/Scripts/UserInterface/EntityCounterViewBehaviour.cs
```diff
@@ -0,0 +1,27 @@
...
+        public void Reset()
+        {
+            _entityCounterLabel.text = "0";
+        }
+
+        private void HandleLivingEntitiesAmountChanged(int amount)
+        {
+            _entityCounterLabel.text = amount.ToString();
+        }
+
```
m4rtint, line 25: u should unsubscribe 
`ifeCycleManager.OnLivingEntityEliminated += HandleLivingEntitiesAmountChanged;`
in the deinit

Assets/_BForBoss/_Entities/Scripts/LifeCycleBehaviour.cs
```diff
@@ -17,6 +18,11 @@ public abstract class LifeCycleBehaviour : MonoBehaviour
 
         public bool IsAlive => _lifeCycle.IsAlive;
 
+        public void NotifyOnDeath(LifeCycleManager manager)
+        {
+            _lifeCycle.NotifyOnDeath(manager);
```
m4rtint, line 23: Not really a review ccomment but just random thoughts. 
The reason why each individual LifeCycle bhaviour/C# object needs to know about the manager. Is because the life cycle behaviour creates their own `new LifeCycle(_health);`

If the manager Initializes each `LifeCycleBehaviour` and DI a `new LifeCycle(_health);` into each one, then you'll be able to subscribe the manager onto each individual entity, instead of having the entity be depenent on a manager.

Gursaanj: Hmm, I can see that. Definitely would like to bring this up on Tuesday too though for discussion

Assets/_BForBoss/_Core/Scripts/UserInterface/EntityCounterViewBehaviour.cs
Gursaanj: by Deinit do you `OnDestroy`?

m4rtint: yuh

Assets/_BForBoss/_Entities/Scripts/LifeCycleManager.cs
```diff
@@ -1,19 +1,23 @@
...
     {
+        public Action<int> OnLivingEntityEliminated;
         private LifeCycleBehaviour[] _lifeCycleBehaviours = null;
-
+        private int _totalEnemiesEliminated = 0;
+        
         public int LivingEntities => _lifeCycleBehaviours.Count(life => life.IsAlive);
 
-        private void Awake()
+        public void EliminateLivingEntity()
```
m4rtint, line 15: hmm so this is `public` because each entity needs to be able to call this right? 
So from an outside perspective, `EleminateLivingEntity()` might give off the meaning of.. Okay this manager is going to kill off a random entity? 

But if you call this, nothing really happens. So it can be misleading.

m4rtint: So what you can do instead. 

Is to pass a lambda function into the entity instead. That way, they don't need to know about the manager, and they just need to run whatever lambda is passed through when they die

m4rtint (changes requested):

Assets/_BForBoss/_Entities/Scripts/LifeCycle.cs
```diff
@@ -46,6 +46,11 @@ public void HealBy(float amount)
             OnHeal?.Invoke();
             _currentHealth = Mathf.Min(_maxHealth, _currentHealth + amount);
         }
+        
+        public void NotifyOnDeath(Action onDeathCallback)
```
m4rtint, line 50: NiT - can the default value of the argument be nil?

Gursaanj: Hmm good call, cant use the `?` operator before Invoke, so will just wrap the parameter in a null check

m4rtint: Can it be set like this? 
`Action onDeathCallback = null`

m4rtint (approved):

PR #142, opened 2022-03-17
BFB-203: Added Panic Helper

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-203

**Description**
Added Panic Helper class
- refactor some of the Debug.LogWarning() 
- Anything that MUST be setup and break the build, I added Panic Helper. 
Some others that are optional, I left alone


unoctium1 (approved):
This looks good, this isn't necessarily a point for this PR, but I do wonder if we should start maybe moving things into the perigon utilities package

m4rtint:
> This looks good, this isn't necessarily a point for this PR, but I do wonder if we should start maybe moving things into the Perigon utilities package

We could. If it'll make things look cleaner. Just not bothered yet. But def would be nice.

PR #143, opened 2022-03-17
BFB-243: Movement Audio Test and slight refactor

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-243

**Description**
- Refactored a little bit of Character movement since Character module + WeaponManager both needed the Character movement details.
- Separated CharacterMovement logic and view

**Screenshots / GIFs**
image


Assets/_BForBoss/Tests/CharacterMovementAudioTests.cs
```diff
@@ -0,0 +1,295 @@
...
+        [Test]
+        public void GetNextStateFromNotPlaying_whenSpeedMagnitudeHigherThanMinSpeed_whenIsWallRunningTrue_resultingStateWallRunning()
+        {
+            // Given
+            var dummyCharacterMovement = new DummyCharacterMovement();
+            var minSpeed = 1;
+            var currentSpeed = 10;
+            dummyCharacterMovement.IsWallRunning = true;
+            dummyCharacterMovement.SpeedMagnitude = currentSpeed;
+            var movementAudio = new CharacterMovementAudio(dummyCharacterMovement, minSpeed);
```
m4rtint, line 20: Basically, these are the state of the character. And the "next state" is the expected result
```
            var minSpeed = 1;
            var currentSpeed = 10;
            dummyCharacterMovement.IsWallRunning = true;
            dummyCharacterMovement.SpeedMagnitude = currentSpeed;
```

Assets/_BForBoss/_Core/Scripts/Managers/GunRangeWorldManager.cs
```diff
@@ -20,7 +20,13 @@ public class GunRangeWorldManager : BaseWorldManager
...
         {
             base.Start();
-            _weaponsManager.Initialize(new CharacterMovementWrapper(_player));
+            _weaponsManager.Initialize(
+                () => _player.CharacterVelocity,
+                () => _player.CharacterMaxSpeed,
+                () => _player.IsWallRunning,
+                () => _player.IsGrounded,
+                () => _player.IsSliding,
+                () => _player.IsDashing);
```
m4rtint, line 29: Basically, instead of a wrapper in a middle of _Core, I'm passing in what WeaponsManager needs.

Originally these values are only needed by weaponsManager, but now it's also needed by Character. So instead of making a copy, i'm moving it back to Character, and just passing in what the other's need. 

This is Dependency inversion.

unoctium1 (approved):
This looks great, getting rid of the wrapper especially is a big improvement imo. All these state machine tests and your messages on slack do make me wonder if we could integrate something into the automated PR messages when PRs include tests to check for how good they are on branch coverage, cause I think that would be useful and isn't the easiest thing to do manually

m4rtint:
> This looks great, getting rid of the wrapper especially is a big improvement imo. All these state machine tests and your messages on slack do make me wonder if we could integrate something into the automated PR messages when PRs include tests to check for how good they are on branch coverage, cause I think that would be useful and isn't the easiest thing to do manually

I know it's somewhat possible. But after a while i find it numbing  (you just start ignoring it).  The most it has does at work is the slight guilt you see. That might be a @mikeslgoh question, but there is a Code Coverage package that I gitignore, which I used

PR #145, opened 2022-03-20
BFB-192: Fix character movement audio bug

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-192

**Description**
Fix bug in character movement audio. Removed event chaining in favor of just passing in one event as a reference. Confirmed to work in build.


m4rtint:
Why is refactoring the solution?

unoctium1:
@m4rtint The previous event chaining seemed to have some weird ordering dependences that worked in editor but not in build. Refactoring just got rid of the chaining. Imo it also seems a little better, cause additional classes similar to `CharacterMovementAudio` should subscribe to the events in `FirstPersonPlayer` rather than the events in the slide or dash behaviours, and so making those events private seems like a bonus

m4rtint (approved):

PR #148, opened 2022-03-23
BFB-259: add code coverage to GitHub PRs

mikeslgoh:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-259


.github/workflows/codeCoverage.yml
```diff
@@ -3,6 +3,7 @@ name: Run code coverage
 on: 
   pull_request: 
     types: [opened]
+  workflow_dispatch: {}
```
mikeslgoh, line 6: this allows the github action to be re-triggered whenever and not just when PRs are created

PR #153, opened 2022-03-31
BFB-263: Lock wall run direction

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-263

**Description**
Lock wall run along wall heading direction. Restructures wall run so that speed is influenced by look direction still, but the player no longer moves where the camera is looking. Velocity is basically projected along wall heading. Also restrict the camera movement so that the camera is only forced along the wall heading once. Once the camera has been stabilized, players are free to look wherever. Also restructure dutch angle so that it looks better, its only actually fully canted when looking down the wall. 

Note: Currently, this works best when the min speed is set to 0, so that players don't fall off the wall when they stop moving. An alternative approach would be to lock player speed to a constant value, but this would prevent changing direction


Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -290,6 +308,18 @@ private Vector3 CalculateWallHeadingDirection()
         {
             return (Vector3.Cross(ChildTransform.up, _lastWallRunNormal) * CalculateWallSideRelativeToPlayer()).normalized;
         }
+        
+        private void LookAlongWall(Vector3 characterForward, Vector3 heading)
+        {
+            if (_hasCameraStabilized) return;
```
m4rtint, line 314: can you make this separate lines

m4rtint (approved):

PR #157, opened 2022-04-01
BFB-264: Added Metallic Height Map Materials

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-264

**Description**

This PR adds in new Materials in `_LevelDesign/Materials/HeightMap` that can be used with the ProBuilder Material Hotkeys. 

### Material Properties
 (The controls for this material are someone involved, so if you aren't used to working with them please let me know and I can explain them directly to you.):

- The created materials have two colour values that can be changed per material. This can be controlled using **Primary Color** and **Secondary Color** properties.

- The created materials have a metallic value and a roughness (called smoothness) value. **Metallic** controls how reflective the material is. **Smoothness** controls how blurry reflections are (for example, mirrors are metallic and smooth, where as  stainless steels fridge are metallic and rough).

- The way normals are calculated from height maps has been improved. You can control how BUMPY the surface is through **Normal Intensity**. **Normal Texel Width** should match the resolution of the height map texture you are using.

- The way textures are randomized has been improved. This is somewhat complex to work with if you haven't touched UVs a lot. I'd recommend playing around with **UV Scale** and **Random Grid Scale** to get a sense of how they interplay. **Use Random Offset** must be on for this to work.

### Creating a New Material
- Duplicate any of the existing materials
- Rename the material
- Change any of the properties

### An Important Note on METALS
How metallic a material looks is entirely based on how it reflects other objects around it. This poses a specific concept to be understood when using metallic materials within Unity Scenes. 

By default. when you apply a metallic material to a mesh in a scene, it will ONLY reflect the skybox. This will make the mesh look... bad for the most part. In the case for URP, we need to use what are called Reflection Probes. 

If you've never Reflection Probes them before, here is a [Brackeys Video](https://www.youtube.com/watch?v=lhELeLnynI8) on it. The key takeaway is, everything you want to reflect should be labelled as **Reflection Probe Static**. This can easily be done by **(1)** making every level piece a child of an empty, **(2)** select the empty parent, **(3)** click on the downward-facing arrow to the right of Static on the top-right on the Inspector, **(4)** click **Reflection Probe Static**, and **(5)** apply to all children. 

Reflection Probe bounding boxes should envelop the geometry to want captured, they also should be placed semi-frequently for a good result.

If you have any results about getting this up and running let me know.


Assets/_BForBoss/_LevelDesign/Materials/HeightMap/M_BlueGreeb.mat
```diff
@@ -0,0 +1,143 @@
+%YAML 1.1
```
m4rtint, line 1: What is greeb?

csweet22: It stands for Greeble which is the patterns that the materials is based on the height map. Could be called anything though really, that just made the most sense to me.

m4rtint:
I think i saw you set up the probuilder material Palette? 

do you think you are able to crate a new Palette. Like 
`BForBoss_Material_Palette` and place whatever you think is suitable? 

So that other developers can just flick through them to see what fits?

csweet22:
> I think i saw you set up the probuilder material Palette?
> 
> do you think you are able to crate a new Palette. Like `BForBoss_Material_Palette` and place whatever you think is suitable?
> 
> So that other developers can just flick through them to see what fits?

I did! But it was causing conflicts with the existing ProBuilder palette so I had deleted it. It also wasn't a particularly good palete. I think it might make more sense for the people who use it to decide what they want in the palete.

csweet22:
> > I think i saw you set up the probuilder material Palette?
> > do you think you are able to crate a new Palette. Like `BForBoss_Material_Palette` and place whatever you think is suitable?
> > So that other developers can just flick through them to see what fits?
> 
> I did! But it was causing conflicts with the existing ProBuilder palette so I had deleted it. It also wasn't a particularly good palette. I think it might make more sense for the people who use it to decide what they want in the palette.

Oh and in addition to that, the palette hotkeys can only work with a few of the keybinds because the higher numbers are bound to other actions and I didn't want to rebind them on my end because I wasn't sure if they were needed or not. 

e.g. Alt+5 conflicts with a Debugger or some sort.

m4rtint:
image


> > > I think i saw you set up the probuilder material Palette?
> > > do you think you are able to crate a new Palette. Like `BForBoss_Material_Palette` and place whatever you think is suitable?
> > > So that other developers can just flick through them to see what fits?
> > 
> > 
> > I did! But it was causing conflicts with the existing ProBuilder palette so I had deleted it. It also wasn't a particularly good palette. I think it might make more sense for the people who use it to decide what they want in the palette.
> 
> Oh and in addition to that, the palette hotkeys can only work with a few of the keybinds because the higher numbers are bound to other actions and I didn't want to rebind them on my end because I wasn't sure if they were needed or not.
> 
> e.g. Alt+5 conflicts with a Debugger or some sort.

Oh yeah, i just ignore Alt+5 in general.

m4rtint:
image
I was wondering, how much more work to get the roughness onto the blocks as well. Or does it have to be per block dependent texturing?

csweet22:
> I was wondering, how much more work to get the roughness onto the blocks as well. Or does it have to be per block dependent texturing?

Could you elaborate on that? Are you asking if we can add a sort of grunge texture to the blocks?

m4rtint:
> > I was wondering, how much more work to get the roughness onto the blocks as well. Or does it have to be per block dependent texturing?
> 
> Could you elaborate on that? Are you asking if we can add a sort of grunge texture to the ProBuilder blocks?

The reference image i put there has e.g. scratches, on the texture. I'm assuming that's the "roughness" texture of the block. 
is it possible to add something generic like that to materials you added?

csweet22:
> > > I was wondering, how much more work to get the roughness onto the blocks as well. Or does it have to be per block dependent texturing?
> > 
> > 
> > Could you elaborate on that? Are you asking if we can add a sort of grunge texture to the ProBuilder blocks?
> 
> The reference image i put there has e.g. scratches, on the texture. I'm assuming that's the "roughness" texture of the block. is it possible to add something generic like that to materials you added?

Just added a new commit with a basic grunge addition to the shader. Let me know how it works for you!

m4rtint:
> > > > I was wondering, how much more work to get the roughness onto the blocks as well. Or does it have to be per block dependent texturing?
> > > 
> > > 
> > > Could you elaborate on that? Are you asking if we can add a sort of grunge texture to the ProBuilder blocks?
> > 
> > 
> > The reference image i put there has e.g. scratches, on the texture. I'm assuming that's the "roughness" texture of the block. is it possible to add something generic like that to materials you added?
> 
> Just added a new commit with a basic grunge addition to the shader. Let me know how it works for you!

Yes nice, this looks natural :)  Thanks

m4rtint (approved):
This LGTM 💯

PR #160, opened 2022-04-01
BFB-199: Changed Dash Speed Lines Implementation

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-199

**Description**
Changed the dash lines from using a particle-based approach to a new shader-based approach.

DashLinesEffect prefab acts as a UI canvas where the dash lines are rendered on. The prefab can be found as a child of the Main Camera. The prefab has a DashLinesEffectBehaviour script that requires a reference to the prefab. The prefab must also be referenced within the player controller. These connections have already been made in the template scenes.

The added shader / material for the dash lines has many options to alter how it looks.

A script was added: Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs


Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,57 @@
...
+        [SerializeField] [Range(0,1)] float _lerpOutStrength = 0.1f;
+        [SerializeField] float _duration = 0.1f;
+        private float _currentTime;
+        
+        [SerializeField] [Range(0,1)] float _maxAlpha = 1f;
+        private Material _speedLineMaterial;
+        
+        private void Start()
+        {
+            _speedLineMaterial = _dashEffect.GetComponent<RawImage>().material;
```
m4rtint, line 29: You don't actually need to setup
`[SerializeField] GameObject _dashEffect;` 
if the component is on the same game object as this script. 

You can just call `GetComponent<RawImage>()` and it'll find it.

csweet22: Good to know! Changed in most recent commit.

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,54 @@
...
+        [SerializeField] float _duration = 0.1f;
+        private float _currentTime;
+        
+        [SerializeField] [Range(0,1)] float _maxAlpha = 1f;
+        private Material _speedLineMaterial;
+        
+        private void Start()
+        {
+            _speedLineMaterial = GetComponent<RawImage>().material;
+            _speedLineMaterial.SetFloat("_alpha",  0);
```
m4rtint, line 27: It's better to put any strings at the top of the class as a `const`. 
`private const string MATERIAL_ALPHA = "_alpha"`   . you can name it better, but something along those lines. 

So you just need to set it once, and if `_alpha` string needs to change, you just need ot change it once.

Assets/_BForBoss/_Character/Scripts/PlayerDashBehaviour.cs
```diff
@@ -12,7 +12,8 @@ namespace Perigon.Character
     public class PlayerDashBehaviour : MonoBehaviour
     {
         [Title("Visual Effects")]
-        [SerializeField] private ParticleSystem _dashVisualEffects = null;
+        [SerializeField] private GameObject _dashEffect = null;
```
m4rtint, line 15: So because you have `private DashLinesEffectBehaviour _dashEffectBehaviour;` below, you can make that a `SerializeField` instead, and pass it in there :) 

And then under `Awake`, check 
```
 if (_dashEffect == null)
            {
                Debug.LogWarning("Dash Effect Missing from PlayerDashBehaviour.cs");
            }
```

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
csweet22: Good point! Changed👍

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+            _speedLineMaterial = GetComponent<RawImage>().material;
+            _speedLineMaterial.SetFloat(MATERIAL_ALPHA,  0);
+            _currentTime = _duration + 10f;
+        }
+
+        private void Update()
+        {
+            var currentAlpha = _speedLineMaterial.GetFloat(MATERIAL_ALPHA);
+            
+            if (_currentTime > _duration)
```
m4rtint, line 37: NiT:
```
var isLerpingToZero = _currentTime > _duration;
var resultingAlpha = isLerpingToZero ? 0 : _maxAlpha

_speedLineMaterial.SetFloat(MATERIAL_ALPHA, Mathf.Lerp(currentAlpha, resultingAlpha , _lerpInStrength))
```

Just to not repeat code.

Assets/_BForBoss/_Character/Scripts/PlayerDashBehaviour.cs
csweet22: Would it make more sense to put the error check under `Awake()` or `SetupVisualEffects()`?

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+        private float _currentTime;
+        
+        [SerializeField] [Range(0,1)] float _maxAlpha = 1f;
+        private Material _speedLineMaterial;
+        
+        private void Start()
+        {
+            _speedLineMaterial = GetComponent<RawImage>().material;
+            _speedLineMaterial.SetFloat(MATERIAL_ALPHA,  0);
+            _currentTime = _duration + 10f;
```
m4rtint, line 30: Can you also set `10f` as a const up top, and name it appropriately. Since other's probably won't know where this magic 10 comes from.

Assets/_BForBoss/_Core/RenderPipeline/UniversalRenderPipelineAsset_Renderer.asset
```diff
@@ -44,6 +44,20 @@ MonoBehaviour:
   m_ShadowTransparentReceive: 1
   m_RenderingMode: 0
   m_AccurateGbufferNormals: 0
+--- !u!114 &2160553602708355395
```
m4rtint, line 47: What are these changes for?

Assets/_BForBoss/_Character/Scripts/PlayerDashBehaviour.cs
csweet22: I definitely didn't realize you could directly pass a GameObject's Script Component instances through Serialize Fields! That is a much clearer solution, thanks for letting me know! :)

m4rtint: Oh I think `Setup Visual Effects`  would be nice.

Assets/_BForBoss/_Core/RenderPipeline/UniversalRenderPipelineAsset_Renderer.asset
csweet22: Oh these are definitely residue from when I was experimenting with the example that you [originally](https://github.com/MirzaBeig/Anime-Speed-Lines) sent. Unity was having some trouble deleting some of the CustomPostProcessing from the pipeline asset. It probably shouldn't have been pushed. I'll delete it 👍

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
csweet22: I've renamed it as `START_TIME_OFFSET`, but the idea isn't easy to communicate through the variable name. Basically the _currentTime has to be an amount (arbitrarily 10) above the _duration to ensure that the dash lines start the scene with an alpha of 0 (at the end of the effect). The key point is that _currentTime is GREATER (>) than _duration rather than just being _duration. Thus it is offset by an arbitrary value.

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+//-----------------------------------------------------------------------
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace Perigon.Character
+{
+    public class DashLinesEffectBehaviour : MonoBehaviour
```
m4rtint, line 14: So below you're trying to `GetComponent<RawImage>()` But there's always that change that there isn't a RawImage on the same GameObject.  Which would then throw a nullpointer in `SetFloat`

However, if you write `[RequireComponent(typeof(RawImage))]` right above the `public class....` line, it should guarantee whoever is using this script that it must have a RawImage. So Unity does the null check for you.

Assets/_BForBoss/_Core/RenderPipeline/UniversalRenderPipelineAsset_Renderer.asset
m4rtint: git command to revert these changes are 
`git checkout develop -- Asset/_BForBoss/_Core/RenderPipeline..../UniversalRenderPipelineAsset_Renderer.asset`

m4rtint:
So i've been experimenting with the dash that you have. and comparing it with the one we previously had. 

I'm wondering if it's possible to make the edges of your changes to be more sharp/clean, and make the faded lines to be solid?

csweet22:
> I'm wondering if it's possible to make the edges of your changes to be more sharp/clean, and make the faded lines to be solid?

I'll look into it!

csweet22:
> > I'm wondering if it's possible to make the edges of your changes to be more sharp/clean, and make the faded lines to be solid?
> 
> I'll look into it!

Having the edges and colour being solid would probably have to be a next iteration as that currently isn't possible with how the shader is working right now.

csweet22:
> git command to revert these changes are git checkout develop -- Asset/_BForBoss/_Core/RenderPipeline..../UniversalRenderPipelineAsset_Renderer.asset

Didn't want this to get lost in the conversation so I'm writing it down here.

I successfully did the git checkout to get the Pipeline Asset from develop, but git is saying my branch hasn't changed (i.e. it says that the new local pipeline asset is the same as the develop branch) so I can't add, commit, or push the pipeline asset to this (BFB-199) branch.

I'm trying to figure out what to do. Do you think I could just delete `UniversalRenderPipelineAsset_Renderer.asset` from the "Files Changed" section in the PR?

csweet22:
Pipeline Render Asset should be reverted to the develop branch's version now! Let me know if it looks alright. 👍

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -0,0 +1,54 @@
...
+        private float _currentTime;
+        
+        [SerializeField] [Range(0,1)] float _maxAlpha = 1f;
+        private Material _speedLineMaterial;
+        
+        private void Start()
+        {
+            _speedLineMaterial = GetComponent<RawImage>().material;
+            _speedLineMaterial.SetFloat(MATERIAL_ALPHA,  0);
+            _currentTime = _duration + START_TIME_OFFSET;
```
m4rtint, line 32: So if it has to be at least 10. 
What you could do, is setup a `[Range(10, something large)]` above the _duration. 

Or name it like `MINIMUM_START_TIME`.

Basically, do you want others to know it has to be 10f minimum, or just force the users without knowing why, that it has to be 10 seconds.

m4rtint: Okay so after looking at this a bit more, i see that you want to make it higher than the duration, So that in the beginning it instantly lerps to 0? 
Instead of having a random dash play on start?

csweet22: > I've renamed it as START_TIME_OFFSET, but the idea isn't easy to communicate through the variable name. Basically the _currentTime has to be an amount (arbitrarily 10) above the _duration to ensure that the dash lines start the scene with an alpha of 0 (at the end of the effect). The key point is that _currentTime is GREATER (>) than _duration rather than being EQUAL to _duration. Thus it is offset by an arbitrary value.

The value just has to be any value above 0 to make sure that the _currentTime starts at the end of the animation. So should I be putting `[Range(0.1, something large)] private const float START_TIME_OFFSET = 0.1f;` ?

csweet22: > Okay so after looking at this a bit more, i see that you want to make it higher than the duration, So that in the beginning it instantly lerps to 0?

Yeah that's the idea!

m4rtint: ahh i see. 
Hmm this might just be a design problem then. 

Let's say you don't count up, but you count down instead.  And then you check if it's `x =< 0`
And then on `Start()`, `_currentTime` would already be at 0.  

Would that remove the need for an `OFFSET`

csweet22: Yeah that's probably a more elegant solution 👍

csweet22:
Alright! I've pushed the change where the timer counts down and there is no start offset. But it looks like some of the connections between the Player controller and the dash effect may have broken in certain scenes. I've fixed it in the Sandbox Template and Racecourse 2.

m4rtint:
> Alright! I've pushed the change where the timer counts down and there is no start offset. But it looks like some of the connections between the Player controller and the dash effect may have broken in certain scenes. I've fixed it in the Sandbox Template and Racecourse 2.

Can you check if there any scenes which may have broken it? is it a configuration issue?

csweet22:
> Can you check if there any scenes which may have broken it? is it a configuration issue?

It's broken in every other scene besides the template and race course 2. It just becomes a question of which scenes should be reconnected because I'm not sure what's worth reconnecting.

m4rtint:
image
Should be okay to replace it anywhere that has this. 

These are the main scenes 
image

csweet22:
> Member

Do you think it would make more sense to just have the DashLines prefab be a part of the PlayerRoot Prefab? That way it just works when you drop a PlayerRoot into any scene?

m4rtint:
> > Should be okay to replace it anywhere that has this. These are the main scenes
> 
> Do you think it would make more sense to just have the DashLines prefab be a part of the PlayerRoot Prefab? That way it just works when you drop a PlayerRoot into any scene?

If it's possible then sure that could work. It used to not be possible because it has to move with the camera

csweet22:
> If it's possible then sure that could work. It used to not be possible because it has to move with the camera

Alright! DashLines prefab has moved to be a child of the PlayerRoot Prefab and I deleted it from every Main Camera in the main scenes.

Let me know how it looks 👍

m4rtint (approved):
This looks good

PR #162, opened 2022-04-07
BFB-199: Fixed _MainTex error in Build

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-199

**Description**
Dash lines were not appearing in build. In debug build, an error was being shown. Dash line shader was changed to removed this error.


m4rtint:
Hmm unfortunately i'm not able to see the dash lines anymore, no errors or anything. But i can't see it in build or within unity.

Even Reimported everything and still can't see it :/

m4rtint:
image
oh and also the lerping out is not used. 

I'm assuming the time should be used based on `isLerpingToZero`

csweet22:
> tried it on MS2 Racecourse

Oh yeah, it wasn't hooked up in there since I thought we were gonna be deleting it. It is now though! Let me know how it looks 👍

m4rtint:
> > oh and also the lerping out is not used. I'm assuming the time should be used based on `isLerpingToZero`
> 
> I removed the `_lerpOutStrength` and renamed the used `_lerpInStrength` to just `_lerpStrength`.
> 
> Could you elaborate on the "I'm assuming the time should be used based on `isLerpingToZero`", are you saying that we should only update `_currentTime` if `isLerpingToZero` is true?

no, i was just making assumptions on how you originally wanted to write your code. 
e.g. 
`var lerpStrength = isLerpingToZero ? _lerpOutStrength : _lerpInStrength`

m4rtint:
image

Ahh, i see that you stuck with this connection. 

Since it's within the same parent, you can apply this specific change to the prefab itself. 
image

m4rtint:
> image
> 
> Ahh, i see that you stuck with this connection.
> 
> Since it's within the same parent, you can apply this specific change to the prefab itself. image

The changes with this will stick to all scenes. Since it's within thesame prefab :)

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
```diff
@@ -38,11 +37,10 @@ private void Update()
...
             var resultingAlpha = isLerpingToZero ? 0 : _maxAlpha;
 
-            _speedLineMaterial.SetFloat(MATERIAL_ALPHA, Mathf.Lerp(currentAlpha, resultingAlpha, _lerpInStrength));
-            
+            _speedLineMaterial.SetFloat(MATERIAL_ALPHA, Mathf.Lerp(currentAlpha, resultingAlpha, _lerpStrength));
+
             _currentTime -= Time.deltaTime;
-            
-            
+
```
m4rtint, line 43: There's a `Debug.Log(_currentTime);` in the Play section, can u remove it.

csweet22:
> The changes with this will stick to all scenes. Since it's within thesame prefab :)

Oh I think it is already working in the prefab! The prefab source shows that the connection to `DashLineBehaviour` looks all good. I think the issue for some reason is that some instances of the player prefab in some scenes have overrides on the DashLineBehaviour that replace it will None. Either way, moving forward it should be good for when you drop a completely un-overrided prefab into a scene.

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
csweet22: Removed!

csweet22:
> > > oh and also the lerping out is not used. I'm assuming the time should be used based on `isLerpingToZero`
> > 
> > 
> > I removed the `_lerpOutStrength` and renamed the used `_lerpInStrength` to just `_lerpStrength`.
> > Could you elaborate on the "I'm assuming the time should be used based on `isLerpingToZero`", are you saying that we should only update `_currentTime` if `isLerpingToZero` is true?
> 
> no, i was just making assumptions on how you originally wanted to write your code. e.g. `var lerpStrength = isLerpingToZero ? _lerpOutStrength : _lerpInStrength`

Oh, I see I see! I've changed it so that lerping in and out uses the same value now. I think it'll be cleaner and clearer that way.

Assets/_BForBoss/_Character/Scripts/DashLinesEffectBehaviour.cs
csweet22: Removed!

csweet22:
So I'm trying to removed the need to have the `DashLineBehaviour` be a serialized reference so that it's less likely to be broken by overrides, but I'm having some trouble getting access to the `DashLineBehaviour` script using code within `PlayerDashBehaviour`

Do you have any idea about how to get the component of a sibling prefab?

image

I can get components from the parent of `PlayerDashBehaviour` (being the PlayerRoot prefab), but from there I don't know how to access the components of the DashLinesEffect Prefab.

m4rtint:
> So I'm trying to removed the need to have the `DashLineBehaviour` be a serialized reference so that it's less likely to be broken by overrides, but I'm having some trouble getting access to the `DashLineBehaviour` script using code within `PlayerDashBehaviour`
> 
> Do you have any idea about how to get the component of a sibling prefab?
> 
> image
> 
> I can get components from the parent of `PlayerDashBehaviour` (being the PlayerRoot prefab), but from there I don't know how to access the components of the DashLinesEffect Prefab from within `PlayerDashBehaviour`.

There's no way to get the sibling value unfortunately. you'lll need basically get the parent, and the `GetComponentFromChildren` .

But I think it's fine to leave it as is.

csweet22:
> There's no way to get the sibling value unfortunately. you'lll need basically get the parent, and the GetComponentFromChildren . But I think it's fine to leave it as is.

Sounds good! So with the changes I've done, is there anything else to do before the PR is good to be merged?

m4rtint:
> > There's no way to get the sibling value unfortunately. you'lll need basically get the parent, and the GetComponentFromChildren . But I think it's fine to leave it as is.
> 
> Sounds good! So with the changes I've done, is there anything else to do before the PR is good to be merged?

Just need to double check some stuff. 
If you are confident everything is okay, you can press `Enable auto-merge (squash)` and it'll auto merge it once it's approvd

m4rtint:
> > The changes with this will stick to all scenes. Since it's within thesame prefab :)
> 
> Oh I think it is already working in the prefab! The prefab source shows that the connection to `DashLineBehaviour` looks all good. I think the issue for some reason is that some instances of the player prefab in some scenes have overrides on the DashLineBehaviour that replace it will None. Either way, moving forward it should be good for when you drop a completely un-overrided prefab into a scene.
> 
> The Blue line make it looks like it's being overrided, but in actuality it shouldn't be. That is just because I had to manually reconnect it in the scene because it was initially overrided by "None". If you revert the override, it should still work, you'd just have to reconnect it to the dash post processing.

so in theory, if all the changes are only in the prefab, you shouldn't need to make any changes to the scene itself. And we try our best not to make unnecessary scene changes to avoid merge conflicts. 

you can revert specific values within a prefab/gameobject. 
So DashBehaviour script, on that specific `SerializedField`, you can revert the dashlines, but not the Post processing. That way it'll keep the connection to both. 

Also you can revert the changes you did to the scene itself. 
git checkout develop -- Assets/_BForBoss/_LevelDesign/Scenes/MS2_RaceCourse.unity

m4rtint (approved):

PR #163, opened 2022-04-12
Fix for ignored metas

unoctium1:
**JIRA Ticket**
*no ticket*

**Description**
Hotfix for ignored meta files. This removes *.meta from the gitignore and adds whatever meta files git picked up after I did that


Assets/UniversalRenderPipelineGlobalSettings.asset.meta
```diff
@@ -0,0 +1,8 @@
+fileFormatVersion: 2
```
m4rtint, line 1: and this file?

unoctium1: Thought it was a file that had been left out by the gitignore change, but its not after all. Just pushed a fix

m4rtint (approved):

PR #170, opened 2022-04-20
BFB-275 : Update to Unity 2021

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-275

**Description**
Update Unity to 2021.3.0f1
Also includes updates to mixpanel and most other packages in use

Note: Due to a urp compatibility issue, there's a regression and dash lines are gone pending an implementation change


ProjectSettings/RiderScriptEditorPersistedState.asset
```diff
@@ -0,0 +1,15 @@
+%YAML 1.1
```
m4rtint, line 1: I don't think this is needed is it?

m4rtint:
Are all the materials changes part of the update?

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
```diff
@@ -48,7 +48,7 @@ public ObjectPooler(string poolName, Func<T> buildObject, Action<T> actionOnGet
 		~ObjectPooler()
         {
 #if UNITY_2021_1_OR_NEWER
-			objectPool.Dispose();
+			objectPool.Clear();
 #else
 			SceneManager.UnloadSceneAsync(poolScene); //destroys all objects in the pool
```
m4rtint, line 53: Shouldn't the rest of the code be removed?

unoctium1:
> Are all the materials changes part of the update?

Yeah, cause URP is updated as well. They aren't meaningfully changed, just reserialized with the new version

Assets/_BForBoss/_Utility/Scripts/ObjectPooler.cs
unoctium1: Sure I'll do that

ProjectSettings/RiderScriptEditorPersistedState.asset
unoctium1: Like ignore it? I can try that, I just as a rule thought new project settings files should be added

mikeslgoh (approved):

PR #172, opened 2022-05-10
BFB-293: Transitioning from URP to HDRP.

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-293

**Description**

This pull request is a LOT of files. It is mostly removing the URP package and adding the HDRP package. Some materials broke and the lighting of most scenes will likely look messed up. Please look through this branch and see how everything looks and if anything in particular needs to be fixed before merging. You may notice some materials that have been replaced, that is because they did not work in HDRP.


m4rtint:
So if you look through the MS2 scene, the bots dying material is messed up.

csweet22:
Should be fixed now

m4rtint:
image
And also this item on the wall

m4rtint (approved):

PR #178, opened 2022-05-26
BFB-278: Add Debug Scene

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-278

**Description**
Adds the Debug Scene (Which currently has the DebugCanvas and LevelDesignFeedbackTool) additively into any scene that has a World manager. This will only affect the Editor and Development scenes.


ProjectSettings/EditorBuildSettings.asset
```diff
@@ -20,4 +20,7 @@ EditorBuildSettings:
   - enabled: 1
     path: Assets/_BForBoss/_LevelDesign/Scenes/CKJ_Weapon_Sandbox.unity
     guid: e09dd7ffe880d1b49939853d42d0020a
+  - enabled: 1
+    path: Assets/_BForBoss/_Core/Scenes/AdditiveDebugScene.unity
+    guid: d05c69bfc85df9741a9cac5511badaab
```
Gursaanj, line 25: I actually don't like this implementation cause it adds the scene for the Prod build, which is wasteful.

Initially wanted to manually add the scene in out BuildPipeline for Development builds via [DevBuildScript.cs](https://github.com/PerigonGames/BForBoss/blob/develop/Assets/Editor/DevBuildScript.cs). But I am not sure how to force a Dev build to try it out and make sure it works

Gursaanj: Any idea @m4rtint @mikeslgoh ?

m4rtint: When you make a manual build in UNity, there's a checkbox that says 
`developer build`

m4rtint: imo, it's not the worse thing to have it in EditorBuildSettings, seeing it should only be one debug Scene, but i understand the concern.

Assets/Editor/ProdBuildScript.cs
```diff
@@ -15,6 +15,12 @@ public static void MyBuild()
         
         foreach (EditorBuildSettingsScene e in EditorBuildSettings.scenes)
         {
+            //Do not include AdditiveDebugScene in ProdBuild
+            if (e.path.Contains("AdditiveDebugScene"))
+            {
+                continue;
+            }
```
Gursaanj, line 22: Turns out it is easier just to remove from our Prod Build pipeline than manually add it to our Dev Build pipeline

Assets/Editor/ProdBuildScript.cs
```diff
@@ -15,6 +15,12 @@ public static void MyBuild()
         
         foreach (EditorBuildSettingsScene e in EditorBuildSettings.scenes)
         {
+            //Do not include AdditiveDebugScene in ProdBuild
+            if (e.path.Contains("AdditiveDebugScene"))
```
m4rtint, line 19: Side thought. 
Debug scene will be the first additively added scene, and we're going to be adding more other ccomponents

Does it mean we'll be naming the other scenes Additive___Scene?

Gursaanj: That's what I was thinking for a naming convention for additive scenes. So that we distinctly know that these scenes will never be the primary scene (so people don't accidentally put something like a world manager or something like that into the scene).

m4rtint (approved):

PR #181, opened 2022-06-09
BFB-215

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-215

**Description**
*[insert here]*

**Screenshots / GIFs**
*[insert here]*


Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -287,26 +379,14 @@ private float GetCameraRoll()
...
-            {
-                _hasCameraStabilized = true;
-            }
-            _baseCharacter.AddYawInput(angleDifference * Time.deltaTime * _lookAlongWallRotationSpeed);
+            direction.y = 0;
+            return Vector3.ProjectOnPlane(direction, _lastWallRunNormal).normalized;
         }
 
-        private static bool GetSmallestRaycastHitIfValid(RaycastHit[] array, out RaycastHit smallest)
+        private bool GetSmallestRaycastHitIfValid(RaycastHit[] array, out RaycastHit smallest)
```
unoctium1, line 389: Is there a reason this isn't static anymore?

m4rtint: no, just don't see a reason to make it non static

unoctium1: Isn't it generally a good idea to use static whenever you can? That was my understanding but I could be mistaken

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -81,50 +114,40 @@ public class PlayerWallRunBehaviour : MonoBehaviour
...
-                else
-                {
-                    Debug.DrawRay(ChildTransform.position, playerFacingDirection * _wallMaxDistance, Color.red);
-                }
-            }
 
-            if (GetSmallestRaycastHitIfValid(hits, out RaycastHit hit) && IsClearOfGround())
+            var shouldStartWallRun = ProcessRaycasts(_startWallRunDirections, out var hit, ChildTransform, _parkourWallMask, _wallMaxDistance)
+                                     && IsClearOfGround()
+                                     && hit.collider != _lastWall;
```
Gursaanj, line 138: Nitpick : Probably could have this long check in its own method `private bool ShouldStartWallRun()`

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -81,50 +114,40 @@ public class PlayerWallRunBehaviour : MonoBehaviour
...
             _baseCharacter = baseCharacter;
             _fpsCharacter = baseCharacter as FirstPersonPlayer;
             _movementInput = getMovementInput;
-            _OnWallRunFinished = OnWallRunFinished;
-            _mask = LayerMask.GetMask("ParkourWall");
+            _OnWallRunFinished = onWallRunFinished;
+            _parkourWallMask = LayerMask.GetMask(PARKOUR_WALL_LAYER);
         }
 
         public void Falling(Vector3 _)
```
Gursaanj, line 131: Wait, what is the point of this parameter? Also if it is needed, can we change its name to something more useful (even if it isn't being used)?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -1,78 +1,111 @@
...
+        [FoldoutGroup("Camera Settings")]
         [SerializeField] 
         private float _lookAlongWallRotationSpeed = 3f;
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField] 
+        private float _minLookAlongWallStabilizationAngle = 5f;
 
-        [SerializeField] private float _minLookAlongWallStabilizationAngle = 5f;
+        [SerializeField]
+        private bool _shouldPrintDebugLogs = false;
```
Gursaanj, line 67: could probably wrap this bool with a 
`#if Unity_Editor || Development_Build`

Since we would rather have it false during a production build! To minimize Debug calls during Production builds

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -322,6 +402,34 @@ private static bool GetSmallestRaycastHitIfValid(RaycastHit[] array, out Raycast
...
+            {
+                Vector3 localDirection = origin.TransformDirection(directions[i]);
+                Physics.Raycast(origin.position, localDirection, out hits[i], maxDistance, layerMask);
+                if (hits[i].collider != null)
+                {
+                    Debug.DrawRay(origin.position, localDirection * hits[i].distance, Color.green);
+                }
+                else
+                {
+                    Debug.DrawRay(origin.position, localDirection * maxDistance, Color.red);
```
Gursaanj, line 419: Could wrap this with the `#if Unity_Editor || Development_Build` preprocessor directive?

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
unoctium1: The parameter doesn't have a purpose, which is why its just a discard character. This is kind of a pattern we've adopted for the custom movement behaviors (sliding, dashing, etc) where ECM2 has a method that we're overriding in `FirstPersonCharacter` and then the override just passes through to some custom logic in one of the custom behaviors. In this case, ECM2's falling state calls this every frame. The parameter is just there cause its a parameter of the character Falling method

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: No not really, it's a case by case. 
https://stackoverflow.com/questions/2671496/when-to-use-static-methods

this kinda sums it up well. 
>Does it make sense to call this method, even if no object has been constructed yet?

As of right now, because it's within an instance of an object, it shouldn't be static, when you extract it out, sure. No need to build for something in the future when it's not used that way right now.

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: Instead of wrapping these everywhere
use
https://answers.unity.com/questions/1301347/how-to-disable-all-logging-on-release-build.html

instead.

Gursaanj (approved):

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
m4rtint: Probably even before that. Like an ApplicationCoordinator of some sort

PR #189, opened 2022-06-19
BFB-330: Auto convert imported base,normal and mask maps into HDRP Materials

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-330

**Description**
When importing a texture file (i.e. with `png` or `tga` extension) that starts with prefix of `B_` (for baseMap), `N_` (for normalMap) or `M_` (for maskMap) into unity, A HDRP material is automatically generated in the same directory with the same name as the given texture (without the prefix).

For creating a HDRP material, a Base Map is the only necessary texture file needed for generation (i.e.it is possible to create a Material importing just a base map, but not possible with a normal and mask map together, without the associate base map with the same name).

The user is able to import as many collections of texture files at once, creating as many Materials as needed.

**Screenshots / GIFs**
AutoGeneratingMaterial
NoProcessingWithoutBaseMap


Gursaanj:
@csweet22  Please let me know if this was the desired outcome and stress test with any edge cases you cant think of. If the case is unlikely to cause any major issue or if we would like some quality of life improvements, we might move that to a second iteration task

csweet22:
I'll test it out! Thanks Gursaanj!

csweet22:
LGTM! I assume Martin will want to look over the actual code portion of the PR, but functionality-wise it looks great to me! Thanks Gursaanj! 👍

m4rtint:
@csweet22 as long as you tested it out, stress tested real situations and your good with it, you can approve it if you want

csweet22 (approved):
LGTM! 👍

PR #195, opened 2022-06-27
Cole/BFB-353: Created Fullscreen Pass Hit Shader

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-353

**Description**

There is a new GameObject in MT_Weapon_Sandbox called HitEffectCustomPassVolume (under Environment). This draws an effect onto the screen. Currently it pulsates when on (and will flash on and then fade off once the health system is working). There is a demo script called HitEffectDemo.cs. This shader controller script is completely placeholder and will be deleted once the health system can control the hit effect.

On the HitEffectCustomPassVolume's HitEffectDemo script, you can controller the speed of the pulse (speed), and how bright the image gets (strength). You can also toggle the effect on and off. 

All new files can be found under Assets > _BForBoss > _Shaders > FullScreenShaders

This PR is the basis for the hit effect. More exposed properties and a different texture can be added in another PR.


Assets/_BForBoss/_Shaders/FullScreenShaders/HitEffectDemo.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+        }
+
+        void Update()
+        {
+            if (isOn)
+            {
+                float factor = (float)( Math.Abs(Math.Sin(Time.realtimeSinceStartup * (1 / speed))) * strength);
+                mat.SetFloat("_EmissionStrength", factor);
```
m4rtint, line 36: What is the range of this factor?

Assets/_BForBoss/_Shaders/FullScreenShaders/HitEffectDemo.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.Rendering.HighDefinition;
+
+namespace BForBoss
+{
+    public class HitEffectDemo : MonoBehaviour
+    {
+        private Material mat;
+        private CustomPassVolume cps;
```
m4rtint, line 12: So are all Post Processing Volume in HDRP `CustomPassVolume`?. 

So even like slow motion/dash volumes would be `CustomPassVolume` instead of `Volume` like we had in URP?

csweet22: Basically yeah. Any fullscreen effect is achieved using a CustomPassVolume in HDRP.

Assets/_BForBoss/_Shaders/FullScreenShaders/HitEffectDemo.cs
csweet22: There is no real limit for the factor.

Assets/_BForBoss/_Shaders/FullScreenShaders/HitEffectDemo.cs
m4rtint: We have a `PostProcessingVolumeWeightTool`, can you see if you can refactor this to make it fit this `CustomPassVolume` usecase?

m4rtint: So URP post processing normally goes between 0 - 1, idk if it'll fit 1:1, but if u want to make something similiar, that could work too.

m4rtint: If it's too much work/out of scope, it's fine. But you should put this script into `Utility`

m4rtint (approved):

PR #196, opened 2022-06-28
Cole/BFB-335: Humanoid Enemy and Flying Enemy

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-335

**Description**
Added in the models, textures, and materials for a Humanoid Robot and Flying Robot.

Added enemies can be seen in PG_Items_Library.

**Screenshots / GIFs**
image


m4rtint:
Do u know if u can add simple rigging to the humanoid one?

csweet22:
I could probably start rigging it yeah! Might make more sense to change up the pipeline for animated assets to include rigging and animations into the asset pipeline. Though I'm not exactly sure how we are animating it since we haven't really figured out procedural vs traditional.

m4rtint:
> I could probably start rigging it yeah! Might make more sense to change up the pipeline for animated assets to include rigging and animations into the asset pipeline. Though I'm not exactly sure how we are animating it since we haven't really figured out procedural vs traditional.

Being that we don't know yet, I think it's better to just go for it and try rigging now. Seeing how humanoids are probably relatively easier and standard.. If we're not sure, just do it now and de

csweet22:
Alright it has a basic rig now!

m4rtint:
1. can you move the enemies under the `Entity` folder? 
2. Not sure if the humanoid  robot is oriented correctly. 
3. can you check if the rigging is setup on Unity. 
image

csweet22:
> 1. can you move the enemies under the Entity folder?

It has now been moved

> 2. Not sure if the humanoid robot is oriented correctly.

I believe that is just a strange artifact from the import. When you drag the model into the scene it should be oriented correctly. In the future I can look into seeing why the preview may look flipped around.

> 3. can you check if the rigging is setup on Unity.

I think it should be working now? (FWIW, The current "rig" is just the actual skeleton/bones without any constraints or controls.)
https://user-images.githubusercontent.com/94721502/176161957-d824f655-c975-4f06-9e68-37b6e90cb767.mp4

m4rtint (approved):

PR #198, opened 2022-06-29
BFB-332 - Integrate custom passes to remove gun camera

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-332

**Description**
Remove gun camera from MS2 Racecourse scene, and replace with custom pass volume. Already integrated into GunRange scene. Note that it seems like this changed up the lighting a bit, as the gun camera was doing some strange things to the lighting on the gun. Gun now appears much brighter


m4rtint:
image


I'm not sure if you're experiencing this, but the weapon itself blurs when u move it around. and also u can see through the weapon.

unoctium1:
Hmm I hadn't noticed that, I'll look into it

PR #201, opened 2022-07-09
BFB-365: AI Shooting at player

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-365

**Description**
Super Simple AI Moving then shooting at player

**Screenshots / GIFs**
https://user-images.githubusercontent.com/5779541/178118121-b72fd3a5-c18d-4297-8942-a807264d1a82.mp4


Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletBehaviour.ApplyDamage.cs
```diff
@@ -1,13 +1,18 @@
-using Perigon.Entities;
 using UnityEngine;
 
 namespace Perigon.Weapons
 {
+    public interface ILifeCycleBehaviour
```
m4rtint, line 5: Because i'm making the Entity depend on the weapon, Unity complains about a cycle dependency 
Weapon Depends on Entity
Entity Depends on Weapon. 

So I can either choose one. So because Weapon could be used in Enemies And Players, I chose Weapon to be the universal one. 

So right now, Entities depend on weapons. So because Weapon is universal, it has this `ILifeCycleBehaviour`. Whoever, or whatever needs this weapon, will need to implement this `ILifeCycleBehaviour`, to use weapons. 

TLDR - Bullet need to cause damage to thing. so as long as the thing needs to weapons, it also needs a life cycle.

m4rtint:
@Gursaanj I moved the Floating Target into _Core

PR #202, opened 2022-07-12
BFB-346: Setup health bar for player

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-346

**Description**
Just used the healthbar behaviour from the enemy

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/178774121-f79cc2eb-e6b5-4faa-9b17-a96c17e24e20.mp4


Gursaanj:
To be honest I think the health bar takes too much away from the view, which is very jarring for an fps, even if it is a health bar!. I think it would be better for a percentage to be used instead (underneath the gun name) as in the crude image below.

takes less space and just as indicative. Bonus points if the number lerps from green to red as the health goes from 100% to 0%

image

m4rtint:
> To be honest I think the health bar takes too much away from the view, which is very jarring for an fps, even if it is a health bar!. I think it would be better for a percentage to be used instead (underneath the gun name) as in the crude image below.
> 
> takes less space and just as indicative. Bonus points if the number lerps from green to red as the health goes from 100% to 0%
> 
> image

image

Gursaanj (approved):

PR #203, opened 2022-07-12
BFB-347: Healing Behaviour

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-347

**Description**
- Simple healing behavior. 

**Screenshots / GIFs**
*[insert here]*


Assets/_BForBoss/_Entities/Scripts/HealingBehaviour.cs
```diff
@@ -0,0 +1,20 @@
+using UnityEngine;
+
+namespace Perigon.Entities
```
Gursaanj, line 3: To be honest, this doesnt seem like an Entity to me, nor general enough to put into our Perigon Namespace.

m4rtint: I put this here because it's related to the lifeCycle, same with Damage being part of entity.

m4rtint: I'm moving this to _core

Gursaanj (approved):

PR #204, opened 2022-07-13
Cole/BFB-299: Created Melee Sparks VFX

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-299

**Description**
Created Melee VFX. A burst of sparks for when melee attacks hit walls or enemies.

The VFX can be found under _BForBoss > _VFX  > MeleeHitEffect

This will need to be integrated with the melee system at some point.

**Screenshots / GIFs**

ezgif com-gif-maker (1)


Gursaanj:
Not necessarily needed for this task, but do you know if the intensity,spread or even color is can be parametrized so they can be changed via code? i.e. let the spread of sparks increase/decrease depending on the speed/intensity of the hit onto the wall!

Gursaanj (approved):

csweet22:
> Not necessarily needed for this task, but do you know if the intensity,spread or even color is can be parametrized so they can be changed via code? i.e. let the spread of sparks increase/decrease depending on the speed/intensity of the hit onto the wall!

Yeah those are definitely something that we can parametrized. It might take a bit of time but I'd be happy to add that to the art backlog! Could definitely make it more versatile.

PR #207, opened 2022-07-16
BFB-355 : Added Enemy Spawner

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-355

**Description**
Added Enemy Spawner (and test scene `EnemySpawner_Test`) that spawns enemies via an ObjectPooler. The time between spawns (instead of instantaneously generating) can be played around via the Enemy Spawner Prefab.

**Screenshots / GIFs**
EnemySpawner


Assets/_BForBoss/_Entities/Scripts/LifeCycleManager.cs
```diff
@@ -9,6 +9,7 @@ public class LifeCycleManager : MonoBehaviour
         public Action<int> OnLivingEntityEliminated;
```
Gursaanj, line 9: General note for this class, since we now have `PlayerLifeBehaviour` which inherits from `LifeBehaviour`. It seems weird the `LifeBehaviorManager` handles the tallying (or event the initialization itself) of Enemies that have been killed. That sort of logic should be handled by some sort of General Enemy Manager class. Which should also handle initializing enemies integrated in the scene and enemies generated from the spawner.

If that makes sense to you @m4rtint , I can go ahead and create a ticket for that (i.e. clean up `LifeCycleManager` and create `EnemyManager`)? It can be handled after #201 gets merged

m4rtint: Yup okay makes sense

m4rtint:
I'm assuming you made a template scene and made a copy of everything?
Screen Shot 2022-07-16 at 6 12 46 AM

m4rtint (approved):
Made a comment but everything else made sense

Gursaanj:
> I'm assuming you made a template scene and made a copy of everything? Screen Shot 2022-07-16 at 6 12 46 AM

Made a scene from the WeaponSandbox Scene Template, but didn't make a new scene template!

PR #212, opened 2022-07-27
BFB-382: Floating AI Animation

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-382

**Description**
Integrated the Animation for Idle, Movement, Shoot

**Screenshots / GIFs**

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Enemies/FloatingEnemyAnimationBehaviour.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Enemies/FloatingEnemyAnimationBehaviour.cs
@@ -1,0 +1,52 @@
+using System;
+using Perigon.Utility;
+using UnityEngine;
+
+namespace BForBoss
+{
+    public interface IFloatingEnemyAnimation
+    {
+        void SetShootingAnimation();
+        void SetIdleAnimation();
+        void SetMovementAnimation();
+    }
+    public class FloatingEnemyAnimationBehaviour : MonoBehaviour, IFloatingEnemyAnimation
+    {
+        private const string ANIMATION_DEATH_KEY = "Death";
+        private const string ANIMATION_MOVEMENT_KEY = "Movement";
+        private const string ANIMATION_SHOOTING_KEY = "Shooting";
+
+        [Resolve] [SerializeField] private Animator _animator = null;
+
+        public void Initialize()
+        {
+            if (_animator == null)
+            {
+                PanicHelper.Panic(new Exception("Animator missing from Floating Enemy Animation Behaviour"));
+            }
+        }
+
+        public void SetMovementAnimation()
+        {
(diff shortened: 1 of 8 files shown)


Assets/_BForBoss/_Core/Scripts/Enemies/FloatingEnemyAnimationBehaviour.cs
```diff
@@ -0,0 +1,52 @@
...
+using UnityEngine;
+
+namespace BForBoss
+{
+    public interface IFloatingEnemyAnimation
+    {
+        void SetShootingAnimation();
+        void SetIdleAnimation();
+        void SetMovementAnimation();
+    }
```
Gursaanj, line 12: I wonder if making `FloatingEnemyAnimation` an abstract/parent class is better than making it an interface?? Cause the methods`SetMovementAnimation`, `SetDeathAnimation`, and `SetIdleAnimation` seems more like consistent logic (that can be overridden if need be) that probably won't require all its inheritors to have unique logic.

i.e. if we have a new floating enemy (called `NewFloatingEnemy`) then `NewFloatingEnemyAnimationBehaviour` will probably look identical to this script.

m4rtint: It's a possibility. The reason for making an interface is for the dependency inversion principle in solid

https://www.digitalocean.com/community/conceptual_articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#dependency-inversion-principle

m4rtint: And don't need to create solution we don't know If we'll use yet. Let's keep it simple for now and update it into abstract if and when we need to.

Gursaanj: hmm, Im okay leaving this as is now, but with future enemies, this will probably lead to a lot of unneeded boiler plate code.

Gursaanj (approved):

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingEnemyAnimationBehaviour.cs
m4rtint: This is specifically made for the FloatingEnemy. You don't have to follow this exact interface. If there's another type of enemy, (ball rolling enemy),  it'll probably have its own IBallRollingEnemyAnimation interface. 

Again, it's not for the reason of polymorphism. It's for Dependency inversion principle in SOLID. 
So that other classes can invert the job of animation to another class. (Within the Shooting script, i shouldn't need a whole animation concrete class, i just need something else to animate shooting for me).

PR #215, opened 2022-08-03
BFB-307 : Convert Weapon Management to an additive scene

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-307

**Description**
Replaces all duplicate instances of weapon systems with an additive weapon manager scene 

Known Issue - the camera follower is jittery. I think this is a cinemachine issue, as directly following the player pivot works fine, but breaks crouching support. Still trying to figure out the best way to deal with this.


Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -40,13 +53,15 @@ protected virtual void Awake()
         {
             _stateManager.OnStateChanged += HandleStateChange;
             _inputSettings = new InputSettings();
+            SceneManager.LoadScene("WeaponManager", LoadSceneMode.Additive);
```
m4rtint, line 56: i'm assuming this is synchronous? 
So Start() can't be called until this load is completed?

Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponAnimationController.cs
```diff
@@ -94,8 +92,11 @@ private void LateUpdate()
         {
             UpdateBobbingWeaponOnMovement();
             UpdateWeaponRecoil();
-
-            _weaponHolder.transform.localPosition = _weaponBobLocalPosition + _weaponRecoilLocalPosition;
+            
+            var camTransform = Camera.main.transform;
```
m4rtint, line 96: Is it better if this is cached? even though it won't solve the issue.

m4rtint:
Screen Shot 2022-08-02 at 11 56 53 PM
Can the weapon camera be added via the AdditiveScene?  And then programaticcally added to the main.camera's child

unoctium1:
> Screen Shot 2022-08-02 at 11 56 53 PM
> 
> Can the weapon camera be added via the AdditiveScene? And then programaticcally added to the main.camera's child

We could, but tbh I think getting rid of the weapon camera is something of a priority. I'd rather just leave it as is for now and work on replacing it with a better system

Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponAnimationController.cs
unoctium1: Yeah should definitely be cached, I'll fix that

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
unoctium1: So it looks like its synchronous but on the next frame, it doesn't seem to currently be causing any issues though

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -22,6 +20,21 @@ public abstract class BaseWorldManager : MonoBehaviour
...
+                if (_weaponSceneManager == null)
+                {
+                    _weaponSceneManager = FindObjectOfType<WeaponSceneManager>();
+                }
+
+                return _weaponSceneManager;
+            }
+        }
+
+        private WeaponSceneManager _weaponSceneManager = null;
```
Gursaanj, line 36: variables above properties

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
Gursaanj, line 56: I think the scene should be called something like `AdditiveWeaponManager` To make it clear that this scene should not be opened by itself! 

That was the convention I was trying to go for with the `AdditiveDebugScene`. What do you think?

m4rtint:
> 

i think it's better to put it in this additive scene right now. so when we do find a solution, it's already isolated in this scene, and any potential solutions is already read to go in this scene.

m4rtint:
> > 
> 
> i think it's better to put it in this additive scene right now. so when we do find a solution, it's already isolated in this scene, and any potential solutions is already read to go in this scene.

So that you don't have to re update all the scenes again

m4rtint (approved):
triedd it, lgtm

PR #222, opened 2022-08-06
BFB-395: Visual Effects Manager

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-395

**Description**
- Singleton Visual Effects Manager

1. Questions - Singleton vs Injecting it in.
2. Should there be a dedicated PlayerVisualEffects Class? - Similiar to Player Audio Class, but this is self contained
3. Should the Visual Effects be handled within each own class, Dash, Health, Slow motion

**Screenshots / GIFs**
Screen Shot 2022-08-05 at 11 03 04 PM


Assets/_BForBoss/_Character/Scripts/PlayerDashBehaviour.cs
```diff
@@ -29,18 +25,16 @@ public class PlayerDashBehaviour : MonoBehaviour
 
         private ECM2.Characters.Character _baseCharacter = null;
         private InputAction _dashInputAction = null;
-        private Action _onDashing = null;
+        private Action _isDashActivated = null;
```
Gursaanj, line 28: Where is this being used? also, the variable is phrased as if it is supposed to be a bool, suggest a name change.

Assets/_BForBoss/_Utility/Scripts/VisualEffectsManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        private const string HEALTH_PASS_EMISSION_KEY = "_EmissionStrength";
+
+        private static VisualEffectsManager _instance = null;
+
+        [Resolve][SerializeField] private CustomPassVolume _healthPassVolume = null;
+        private CustomPassVolumeWeightTool _healthVFXTool;
+
+        public static VisualEffectsManager Instance => _instance;
+
+        public void Awake()
```
Gursaanj, line 25: does this need to be public?

m4rtint: no. will change

Assets/_BForBoss/_Character/Scripts/PlayerDashBehaviour.cs
m4rtint: i'll revert it.

Gursaanj (approved):

PR #223, opened 2022-08-07
BFB-389 - Flying Enemy Knockback

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-389

**Description**
Adds flying enemy specific knockback support. Now, their navmeshes temporarily turn off and they fall out of the sky. They also spin out, this can be fixed if we want by limiting the angular velocity on the rigidbody but I thought it was a kinda cool unintended effect. 

**Screenshots / GIFs**
https://youtu.be/5mCU5efehDs


BForBoss.sln.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
Gursaanj, line 1: What is this file for?

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
```diff
@@ -63,15 +73,49 @@ protected override void LifeCycleFinished()
...
                 Debug.LogWarning("A FloatingTargetBehaviour is missing a health bar");
             }
         }
+
+        private void CheckKnockback(float time)
+        {
+            if (_rb == null) 
+                return;
+            _knockbackCurrentTime += time;
+            if (_knockbackCurrentTime > _knockbackMaxDuration || _rb.velocity.sqrMagnitude < float.Epsilon * float.Epsilon)
```
Gursaanj, line 98: interesting, why `float.Epsilon * float.Epsilon` ?

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
```diff
@@ -83,9 +127,19 @@ private void FixedUpdate()
...
                     break;
+                case FloatingTargetState.KnockedBack:
+                    CheckKnockback(Time.fixedDeltaTime);
+                    break;
             }
         }
-        
 
+
+        public void ApplyKnockback(float force, Vector3 originPosition)
```
Gursaanj, line 137: public methods should be above private ones

Assets/_BForBoss/_Entities/Scripts/Navigation/AgentNavigationBehaviour.cs
```diff
@@ -41,11 +40,33 @@ public void MovementUpdate()
...
+                }
+            }
+        }
+
         private void Awake()
         {
             _agent = GetComponent<NavMeshAgent>();
             if (_agent == null)
             {
+                //Do we need this? the component has a RequireComponent flag on it already
```
Gursaanj, line 69: Fair point, maybe this was added just in case the `RequireComponent` attribute gets removed in the future? I'm down for removing this too though

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
```diff
@@ -63,15 +73,49 @@ protected override void LifeCycleFinished()
             {
                 gameObject.SetActive(false);
             }
+            ToggleKnockbackState(false);
         }
 
         private void Awake()
         {
+            _rb = GetComponent<Rigidbody>();
```
Gursaanj, line 81: you could lazy initialized `_rb` (i.e. if `_rb == null` {_rb == AddComponent<Rigidbody>();}) . Then you can remove all the subsequent `if (_rb == null) return` in the other methods

Gursaanj (commented):
Looks awesome!!

Assets/_BForBoss/_Entities/Scripts/Navigation/AgentNavigationBehaviour.cs
m4rtint: Just remove it. Overlooked code

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
m4rtint: Then a require component rigid body should be added then.

If it's not required, then you need null checks and a log warning.

m4rtint:
Btw as cool as it is for it to fall like that, it'll probably also have a lot of side effects. E.g. What happens while it's shooting?.
Since this is basically a stun lock when it falls. (or at least that is what it would look like)

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
unoctium1: using squared magnitude and comparing against a squared value is slightly more performant than comparing against magnitude itself. Since epsilon is just a small value placeholder I don't know that it matters that much though

BForBoss.sln.DotSettings
unoctium1: Oh yeah, I explained that in the commit but not the pr description, whoops! It just tells Rider (and I think visual studio? not sure about that though) that Knockback isn't actually a typo, cause the constant warnings were annoying me. If you guys prefer I can undo that and leave it at the user settings level, but I figure there was no harm in making it project wide, since we're using Knockback as a term all over the place.

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
unoctium1: Yeah my thinking was that we should allow for there not to be a rigidbody, and just have knockback do nothing if thats the case. I'm okay with making it `RequireComponent` though

unoctium1:
> Btw as cool as it is for it to fall like that, it'll probably also have a lot of side effects. E.g. What happens while it's shooting?. Since this is basically a stun lock when it falls. (or at least that is what it would look like)

I'm not sure what we can do instead of it falling like that though? Thats just an effect of having a rigidbody with gravity, the only way for it not to fall would be to drastically ramp up the force of knockback so that its more launched than it currently is, and that would have the same amount of side effects. Regarding shooting though, since Knockback is implemented as a state I don't think there is a way for it to interfere with shooting

m4rtint:
I'm good whenever

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
```diff
@@ -6,18 +6,28 @@
...
     [RequireComponent(typeof(AgentNavigationBehaviour))]
-    public class FloatingTargetBehaviour : EnemyBehaviour
+    public class FloatingTargetBehaviour : EnemyBehaviour, IKnockback
     {
         [SerializeField] private HealthbarViewBehaviour _healthbar;
+
+        [Header("Knockback Settings")]
+        [SerializeField] private float _knockbackRadiusModifier = 10f;
+        [SerializeField] private float _knockbackUpwardsForceModifier = 10f;
+        [SerializeField] private float _knockbackMaxDuration = 2f;
```
m4rtint, line 16: Don't have to do it now, but if you feel like the `FloatingTargetBehaviour` is too bloated with different things, this could be abstract out into a separate knockbackBehaviour class component,  feel free to do so.

m4rtint:
> > Btw as cool as it is for it to fall like that, it'll probably also have a lot of side effects. E.g. What happens while it's shooting?. Since this is basically a stun lock when it falls. (or at least that is what it would look like)
> 
> I'm not sure what we can do instead of it falling like that though? Thats just an effect of having a rigidbody with gravity, the only way for it not to fall would be to drastically ramp up the force of knockback so that its more launched than it currently is, and that would have the same amount of side effects. Regarding shooting though, since Knockback is implemented as a state I don't think there is a way for it to interfere with shooting.
> 
> Edit: Would you prefer having the same implementation, but leaving gravity off so its more like floating in space? I can see that potentially working well

you could try it and see if it works better. 

Another work around i could potentially see, is that it teleports from the floor back up (using the spawn in vfx). so it's moree of a transition/visual feedback rather than just suddenly baack up.

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
Gursaanj: Fair enough, could declare it as a Const/Readonly variable? to avoid doing the multiplication each time

Assets/_BForBoss/_Core/Scripts/Enemies/FloatingTargetBehaviour.cs
unoctium1: Yeah I like your idea of having VFX trigger on knockback end and I think that will really start to bloat this. I'll leave that feature itself for later but pull this out for now

Gursaanj (approved):

m4rtint (approved):

PR #227, opened 2022-08-09
BFB-390: Created Slomo Post Process Volume

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-390

**Description**
Added an HDRP post processing volume for a slomo effect to the PostProcessingVolume prefab.

This effect will need to be triggered by the player controller through code. When pressing slomo button the weight of the volume should slowly interpolate from 0 to 1. When released the weight should interpolate back to 0.

**Screenshots / GIFs**
image


Gursaanj (approved):
Looks Awesome

PR #233, opened 2022-08-15
Physics Layer Fixes

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-411
https://perigongames.atlassian.net/browse/BFB-404

**Description**
Creates a number of new layers, and integrates into weapon systems. There is now a Player layer, an enemy layer, an Affects Player layer (to be used for things like the death plane, that should only ever be colliding with the player) and an EnemyBullet layer (a copy of the bullet layer, but only able to damage the player). Also integrates the layers into the bullet spawner and the melee system.

Also adds an attribute to select a layer, cause apparently that isn't built into unity


ProjectSettings/TagManager.asset
```diff
@@ -16,10 +16,10 @@ TagManager:
   - Bullet
   - FirstPersonWeapon
   - TriggerArea
-  - 
-  - 
-  - 
-  - 
+  - Player
+  - AffectsPlayer
```
m4rtint, line 20: Why do we need this `AffectsPlayer` layer? as opposed to just fixing the logic in Deathplane?

Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletSpawner.cs
```diff
@@ -8,6 +8,8 @@ public class BulletSpawner : MonoBehaviour
         [SerializeField] private BulletBehaviour[] _bulletPrefabs;
         private ObjectPooler<BulletBehaviour>[] _pools = null;
 
+        [SerializeField, Layer] private int _layer = 8; //default is Bullet layer
```
m4rtint, line 11: Would it be more accurate to find the layer via string? since sorting and changing the layers would affect this

ProjectSettings/TagManager.asset
unoctium1: I guess we don't strictly need it? Both approaches feel like they have their pros and cons. The benefit in my opinion is that if we want to extend this behavior to something else (ie. the health pickups) its just a matter of changing a prefab's layer

Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletSpawner.cs
unoctium1: Yeah I was on the fence about this. The downside of having a string is that it plays less well with the inspector, with an int its just a matter of selecting the layer in a dropdown. With a string, you're right its a little more robust to later changes to the layer order, but initially setting it is more of a process, and you have to handle converting to an int at runtime anyway

ProjectSettings/TagManager.asset
m4rtint: Also good to  remember, somebody else might come in and try to implement picking up, but it won't be obvious to them why it's not working until they either figure out it's something in the layer vs their logic.

m4rtint (approved):

ProjectSettings/TagManager.asset
unoctium1: Yeah I'm open to changing it if you prefer, up to you

m4rtint: yeah can u change it?

m4rtint (changes requested):
Removing Player allow layer
- using logic to determine death layer

Assets/_BForBoss/_Utility/Scripts/PropertyAttributes/LayerAttribute.cs
```diff
@@ -0,0 +1,14 @@
+using System.Collections;
+using System.Collections.Generic;
```
Gursaanj, line 2: Can probably remove these

Assets/_BForBoss/_Weapons/Scripts/Weapons/MeleeWeapon.cs
```diff
@@ -16,6 +16,8 @@ public class MeleeWeapon
         private Action<bool> _onHitEntity;
         public bool CanMelee => _currentCooldown <= 0f;
         private int _hits;
+
+        private int layerMask = ~LayerMask.GetMask("Player");
```
Gursaanj, line 20: So we have this [class](https://github.com/PerigonGames/BForBoss/blob/6a287ac5b6b9127267d484e514ddc8133b678a2d/Assets/_BForBoss/_Utility/Scripts/Tags.cs#L12) that is technically meant to keeps all the layer and tags string, to reduce typing errors with these names. 

Not sure if you would rather use that?? I am down either way!

m4rtint: yeah should put that in there.

ProjectSettings/TagManager.asset
unoctium1: I think after thinking about this further it might be better to keep the layer approach - string comparisons just seem so much messier and though I get its minimal, I could see the performance benefit from using Unity's built in physics helping later down the line. In addition if we decide death planes shouldn't be visible that becomes a super small fix

m4rtint (approved):

PR #235, opened 2022-08-15
BFB-394: Wall running Fix

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-394

**Description**
Checks if the angle between last wall and next wall more than {70} variable

**Screenshots / GIFs**


Assets/_BForBoss/_LevelDesign/Scenes/WaveSandbox.unity
```diff
@@ -5321,82 +5320,6 @@ Mesh:
     offset: 0
     size: 0
     path: 
---- !u!1 &315269722
```
m4rtint, line 5324: I just removed the Event System from the map

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -307,6 +314,12 @@ private bool IsTooFarFromWall()
             
             return true;
         }
+
+        private bool IsNextWallAngleTooObtuse()
+        {
+            var angleDifference = Vector3.SignedAngle(ChildTransform.forward, _lastPlayerWallRunDirection, Vector3.up);
+            return angleDifference == 0;
```
Gursaanj, line 321: since `angleDifference` returns a float, maybe we want to use `Mathf.Approximately(angleDifference, 0.0f)` to take into account floating pointers (https://docs.unity3d.com/ScriptReference/Mathf.Approximately.html).

Assets/_BForBoss/_Character/Scripts/PlayerWallRunBehaviour.cs
```diff
@@ -307,6 +314,12 @@ private bool IsTooFarFromWall()
             
             return true;
         }
+
+        private bool IsNextWallAngleTooObtuse()
```
Gursaanj, line 318: Nitpick : So Obtuse means an angle over 90 degrees. `TooObtuse` sounds kinda vague/arbitrary (kinda like how obtuse is considered too obtuse). So I wonder if we should think of another name for the method.

Note that I can't think of anything at the top of my head at the moment, so I'm okay with this for now. @unoctium1 any ideas??

unoctium1: Yeah I'm not sure I really understand what this is doing, isn't it just returning true if the player forward direction and the wall run direction are the same?

m4rtint: it returns the angle between the last *hit* normal and direction.  
I was experimenting with the angles and this being 0 seems to work. Please feel free to mess around with it.

unoctium1: Are you sure? I thought `_lastPlayerWallRunDirection` is the player's direction along the wall, not the normal. To me this looks like its returning true when moving in a straight line along a wall

m4rtint:
Completely changed the logic to 
1. Have an obtuse angle 
2. Checks last wall run direction vs the next wall run direction. If the angle between them is larger than 70 (Default value), then it'll stop wall running

Gursaanj (approved):

PR #238, opened 2022-08-18
Cole: Level Design Tests

csweet22:
Cole's Level Design Tests!


m4rtint:
Just an fyi, i'm making changes to the weapons + animations. so it'll probably mean a lot of conflicts in the additive weapon scene.

One requirement i have in mind, is to change the Scout Rifle into a pistol (like a hard hitting pistol).

PR #240, opened 2022-08-19
Add mesh link cost dirty

Gursaanj:
Added an impromptu (quick and dirty- do not merge into main) way for Navmesh agents you use side by side navmesh offmesh links instead of making them wait in a line before traversing 

My changes add a test section with 3 side by side offmesh links with the same endpoint transform. 

This is achieved by raising the cost of the currently used offmeshlink when in use and setting it back to the original value when the agent leaves the link

**Screenshots / GIFs**
Multiple Off Mesh Links
Three In ONe
Temp Checker script
Navigation


Gursaanj:
I merged with my branch to get the UI working and stuff, so there might be quite a few conflicts. I suggest just copying over the script and adding it to your prefab (and re-delete the dummy prefab from the scene, if it still there). Then just let me know and I can go ahead and delete this branch

Assets/_BForBoss/_Core/Scripts/Environment/OffMeshLinkChecker.cs
```diff
@@ -0,0 +1,63 @@
+using UnityEngine;
+using UnityEngine.AI;
+
+namespace BForBoss
+{
+    public class OffMeshLinkChecker : MonoBehaviour
+    {
+        [SerializeField] private NavMeshAgent _agent = null;
+        [SerializeField, Tooltip("How much should the cost of a mesh link multiply by when in use?")]
+        private float _multiplier = 1000f;
```
m4rtint, line 10: Does this need to be dynamic variable? or just some random high value?

m4rtint: I guess this would work if you want enemies to come in a certain order.

Gursaanj: yeah the high value is to discourage the enemies from waiting for an enemy to be done using this offmeshlink instead of trying to find a different path with a lower cost.

csweet22:
Sorry for the delay! I've added it to my test branch and it seems to be working! Should be good to delete this PR now. Thank you Gursaanj!

PR #242, opened 2022-08-24
BFB-388 : Enemy Shooting VFX Integration

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-388

**Description**
- Refactor enemy muzzle vfx to use a custom event for firing, and remove the Shoot Delay property - now it is up to the controlling script to call OnFire when muzzle flash should be triggered
- Create enemy bullet prefab
- Update reference in entity manager bullet pool so that correct prefab is used
- Add vfx to enemy shooting behaviour

**Screenshots / GIFs**
https://youtu.be/2ybWrQNsh1A


unoctium1:
@csweet22 Adding you as a reviewer mostly to get your feedback on timings, VFX size/placement, etc.

csweet22:
Can / should we delete this child in the prefab? I think that it isn't actually doing anything so it might be a good idea to get rid of it.
image

csweet22:
Could we also bring the scale of the charging effect found in the FloatingEnemy prefab from 0.5 to 0.1?

csweet22:
Placement wise can we move the ShootingFrom object from (0, 2, 0.621) to (0, 2.1, 0.5)? I think that that might look better with the smaller scale suggestion.

csweet22:
> Placement wise can we move the ShootingFrom object from (0, 2, 0.621) to (0, 2.1, 0.5)? I think that that might look better with the smaller scale suggestion.

I'm also wondering if it would make sense to have ShootingFrom be a child of the Root of the actual Mesh's Armature because when the enemy animates the charge orb isn't moving with it so it gets misaligned.

csweet22:
And could we also change the Charge Time of the P_EnemyMuzzle found in the FloatingEnemy prefab to 1.5 and the muzzle time to 0.2? I find that that feels nicer.

unoctium1:
> And could we also change the Charge Time of the P_EnemyMuzzle found in the FloatingEnemy prefab to 1.5 and the muzzle time to 0.2? I find that that feels nicer.
> 
> image

The muzzle duration can be changed, the charge time should be dynamic now though, where it actually uses `_shootCountDownInSeconds` from the enemy shooting AI. If you think its too short though, I could either extend that value or use both it and the aiming count down, which would bring it up to about 1s up from 0.5

unoctium1:
> > Placement wise can we move the ShootingFrom object from (0, 2, 0.621) to (0, 2.1, 0.5)? I think that that might look better with the smaller scale suggestion.
> 
> I'm also wondering if it would make sense to have ShootingFrom be a child of the Root of the Mesh because when the enemy animates, the charge orb isn't moving with it so it gets misaligned.
> 
> If it isn't the best practice to make the bullets fire from a child of the mesh itself, it might be a good idea to just have another ShootingFrom point just for the VFX to be aligned with the mesh as it animates.
> 
> (Surprisingly, the root of the armature works perfectly as the firing point.)
> 
> image

@m4rtint Do you see any concerns with this? I can certainly change it, just want to check it won't break the enemy animation stuff you set up

m4rtint:
Feel free to set it up.  The animation is mostly from magan on blender

csweet22:
> @csweet22 The muzzle duration can be changed, the charge time should be dynamic now though, where it actually uses _shootCountDownInSeconds from the enemy shooting AI. If you think its too short though, I could either extend that value or use both it and the aiming count down, which would bring it up to about 1s up from 0.5

Oh that's completely my bad I had read the code wrong! 

I think that it makes sense to just use the value of `_shootCountDownInSeconds` and not have to add in the aiming count down. 

I think that visually and gameplay-wise it would be nicer to have the `_shootCountDownInSeconds ` to be a higher value like 1.5s, but at the end of the day it will depend on what whoever wants to decide game design decisions thinks is best.

Everything looks great though, thank you for integrating the effect! 🙌

m4rtint (approved):

PR #254, opened 2022-09-08
BFB-428: Remove Additive Scenes from Scene Switcher Debug Window

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-428

**Description**
Removed additive scenes from showing up in the SceneSwitcher Debug Window by not adding scenes that start with "Additive" (Which is the convention we have softly adopted when creating additive scenes). Though not perfect as it is for the Debug menu it should be fine

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Debug/DebugView/SceneSwitcherDebugView.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Debug/DebugView/SceneSwitcherDebugView.cs
@@ -79,12 +79,10 @@ namespace BForBoss
             {
                 string buildSceneName = GetAppropriateSceneName(SceneUtility.GetScenePathByBuildIndex(i));
 
-                if (buildSceneName.StartsWith(ADDITIVE_SCENE_PREFIX))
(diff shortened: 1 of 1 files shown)


Assets/_BForBoss/_Core/Scripts/Debug/DebugView/SceneSwitcherDebugView.cs
```diff
@@ -76,7 +77,14 @@ private void GetBuildSceneNames()
 
             for (int i = 0; i < SceneManager.sceneCountInBuildSettings; i++)
             {
-                _buildSceneNames.Add(GetAppropriateSceneName(SceneUtility.GetScenePathByBuildIndex(i)));
+                string buildSceneName = GetAppropriateSceneName(SceneUtility.GetScenePathByBuildIndex(i));
+
+                if (buildSceneName.StartsWith(ADDITIVE_SCENE_PREFIX))
```
m4rtint, line 82: Can you reverse this? Continue is a bit unnatural

m4rtint (approved):
Can update if you want

PR #258, opened 2022-09-11
BFB 425: Enemies spawn at their designated Spawn Location instead of 0,0,0

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-425

**Description**
Due to an internal bug with Unity `NavMeshAgent` we have to do a cheap trick to ensure that agents spawn appropriately at runtime. In which, we just have to disable/enable the enemy's `NavMeshAgent` after Instantiating it and setting its transform.

Reference used:
https://answers.unity.com/questions/771908/navmesh-issue-with-spawning-players.html

**Screenshots / GIFs**
PR attachment

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
@@ -1,6 +1,8 @@
 using System.Collections;
 using Perigon.Entities;
 using UnityEngine;
+using UnityEngine.AI;
+
 namespace BForBoss
 {
     public class EnemySpawnAreaBehaviour : MonoBehaviour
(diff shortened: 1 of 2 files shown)


Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
```diff
@@ -84,8 +86,16 @@ private void SpawnEnemy()
...
             var enemy = _enemyContainer.GetEnemy();
+            
             enemy.OnRelease += HandleOnEnemyReleased;
             enemy.transform.SetPositionAndRotation(transform.position, transform.rotation);
+            enemy.transform.SetParent(transform, true);
+            
+            //https://answers.unity.com/questions/771908/navmesh-issue-with-spawning-players.html
+            NavMeshAgent agent = enemy.GetComponent<NavMeshAgent>();
+            agent.enabled = false;
+            agent.enabled = true;
```
m4rtint, line 97: can u not do something like this within the `AgentNavigationBehaviour`? 

Or create a `Spawn` State in the AI?

Gursaanj: I tested quite a bit with enabling and disabling the agent in `AgentNavigationBehaviour` but the results seem inconsistent or hard to follow!

It looks like only after we Set the transform of the Instantiated EnemySpawn do we have to disable/enable the agent for the most consistent results

m4rtint (approved):

PR #262, opened 2022-09-13
BFB-432: AI Logic Updated

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-432

**Description**
- Added Distance and Object obstruction checks. 

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/189952175-65fb4754-269b-4fc0-8ec4-daa4525e93b7.mp4


m4rtint:
@Gursaanj Did a slight refactor. 
`IsLineOfSight` is its own class,  which is passed in. Since it'll probably be used universally. 
And moved the navigation agent into the _Core Folder, seems more accurate to group all AI components into core.

Assets/_BForBoss/_Core/Scripts/Enemies/AgentNavigationBehaviour.cs
```diff
@@ -2,21 +2,23 @@
 using UnityEngine;
 using UnityEngine.AI;
 
-namespace Perigon.Entities
+namespace BForBoss
 {
     [RequireComponent(typeof(NavMeshAgent))]
     public class AgentNavigationBehaviour : MonoBehaviour
     {
         [SerializeField]
```
Gursaanj, line 10: if we want to be safe, we can make this field have a `Min` attribute so that this value can't be less than 0.

Assets/_BForBoss/_Core/Scripts/Enemies/AgentNavigationBehaviour.cs
```diff
@@ -61,16 +67,27 @@ public void ResumeNavigation()
...
         }
 
+        private bool ReachedDestination()
+        {
+            if (_agent.remainingDistance > 0)
+            {
+                return _agent.remainingDistance < _stopDistanceBeforeReachingDestination;
+            }
+
+            return false;
```
Gursaanj, line 77: Can rewrite this to `return _agent.remainingDistacnce > 0.0f && _agent.remaningDistance < _stopDistanceBeforeReachingDestination;`

Assets/_BForBoss/_Core/Scripts/Enemies/IsLineOfSightBlocked.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using Perigon.Utility;
+using UnityEngine;
+
+namespace BForBoss
+{
+    public class IsLineOfSightBlocked
```
Gursaanj, line 7: Don't think this is necessary. Could probably make this a static helper class (i.e. `public static class LineOfSight`) and instead of `Execute` make it `public bool IsBlocked(Vector3 destination, Vector3 shootFromPosition){}`.

Passing this class as a variable seems like it adds no benefit atm since no information is being stored/kept.

Gursaanj: or since everything is now under the `BForBoss` namespace, you could make it `internal static class LineOfSight'

m4rtint: It needs to store the `Func<Vector3>` not just Vector3. Every enemy has their own `Func<Vector3>` that constantly changes. 

If I were to make this a static class, then
1. bad for testing.
2. Then it'll be another callback from the shooting/navigation classes.

m4rtint: I can change the name though

Gursaanj: Hey, sorry for the late reply. Honestly, I'm still not 100% convinced that this can't just be a helper class.

Like `EnemyShootingBehaviour` has both `getPlayerPosition` and `shootingFromPosition` so `LineOfSight' has no need for having its information.

And for `AgentNavigationBehaviour` has `navigationDesitnation`. So feasibly we could pass in Another Func<Vector3> (i.e. `getPlayerPosition`) into the Initialize Function instead of `LineOfSight` and instead Call the `LineOfSight` static helper class.

Let's talk about this during the Work session today!

m4rtint: The reason of having a separate class handle it is 
1. reusability.
2. Doesn't break SOLID principle (single responsibility principle). The job of  floating target is for connecting everything, having a random helper function there would signal the fact that all shared logic should be placed in the parent class.

If we really want to break reason 1, sure since we only have one ai for now.

If we want to break 2, then we'll have start having callbacks for any shared code between components in their initializes as a pattern. Which we have been doing, but also had to start refactoring since it went too far.


I'll leave this decision up to you if I can't change your mind after this.

m4rtint: And havin LineOfSight makes it explicit as well, so that we know we need this explicit logic as opposed to. A generic function to pass in that could be any logic.

It's like when setting up a class initialize, it forces another dev to use line of sight instead of passing in a random function.

Gursaanj (approved):

PR #263, opened 2022-09-14
BFB-415: Created Fade-to-Black Pass 

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-415

**Description**
Created a CustomVolumePass called FadeToBlackCustomVolumePass in the PostProcessingVolume prefab.

I changed what-I-think-to-be the appropriate code in the Visual Effects Manager script to change the death Volume to type CustomVolumePass. (This may have been done incorrectly, so please let me know if that is the case.)

As per usual, the intensity of this custom pass can be controlled by a single value. 0 means no black, 1 means full black.

NOTE: Because of how CustomVolumePasses work, I was unable to find a way to also blackout the UI and the weapon. Further investigations may need to be done to figure out the best way go about having the UI and weapon not show up on a fade-to-black as I am not 100% sure how the UI and Weapon are being renderer currently.

### DIFF
--- a/Assets/_BForBoss/_Core/Prefabs/PostProcessingVolume.prefab
+++ b/Assets/_BForBoss/_Core/Prefabs/PostProcessingVolume.prefab
@@ -99,7 +99,7 @@ Transform:
   m_ConstrainProportionsScale: 0
   m_Children: []
   m_Father: {fileID: 2228368853135509604}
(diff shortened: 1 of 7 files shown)


m4rtint:
You were just missing, setting up the camera

m4rtint (approved):

PR #266, opened 2022-09-14
BFB-424 : Debug Log Controller

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-424

**Description**
Add custom logging
 - Globally enablable/disablable via Tools > Use Logging
 - Configurable on a per key basis via Preferences > Perigon > Debug Logging

_**Changed to use keys instead of namespaces**_
Some notes on this:
- Keys are stored per project (See the new Perigon/Logger.asset file), while if they are enabled or not is a per user setting. 
- At runtime, there is no functional difference between a key not existing or a key not being enabled, except for a logged warning. I thought about using enums and codegen to prevent that, but the downside seemed to outweigh benefits. I think logging a warning is probably enough, but would be interested in hearing other opinions. 

**Screenshots / GIFs**
image


m4rtint:
Can you explain Logger file vs Logger Editor only file?

Assets/_BForBoss/_Core/Scripts/Managers/WaveManager.cs
```diff
@@ -36,7 +37,7 @@ private void OnEnemySpawned()
         {
             if (_waveModel.IsMaxEnemySpawnedReached)
             {
-                Debug.Log("<color=Blue>Max Enemy Reached</color>");
+                Logger.LogString("<color=Blue>Max Enemy Reached</color>");
```
m4rtint, line 40: How do you differentiate between the different logs in different places? 
There are logs in WaveModel as well, which i would group together with this.

unoctium1: Ah I missed WaveModel, should have updated it now though. Scripts are grouped by namespace, so Waves will all be under the root BForBoss toggle at the moment

unoctium1:
> Can you explain Logger file vs Logger Editor only file?

@m4rtint its just a convenience thing, to prevent having editor compilation flags all over the place. Logger contains methods to be called at runtime, and EditorOnly is entirely wrapped in editor flags.

Assets/_BForBoss/_Core/Scripts/Managers/WaveManager.cs
m4rtint: Is there a way to group it by a Key that we pass in? instead of namespaces?
`Logger.LogString("<color=Blue>Max Enemy Reached</color>", "Key");`

unoctium1: I guess if we hard code in preset keys? I was initially thinking we could do something like use reflection to find all the keys used across the project and build the editor that way, but it seems to be impossible without using like roslyn analyzers.

unoctium1: It would be a lot messier than the current approach, but having a singleton scriptable object that acted as a key repository could work

m4rtint:
i'm not able to create a build on my local machine curently.

m4rtint (approved):

PR #270, opened 2022-09-20
BugFix/Prod build fix

m4rtint:


Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -109,6 +114,19 @@ protected virtual void OnValidate()
             }
         }
         
+        private void OnAdditiveSceneLoaded(Scene scene, LoadSceneMode loadSceneMode)
+        {
+            if (scene.name == ADDITIVE_USERINTERFACE_SCENE_NAME)
```
Gursaanj, line 119: can do a switch statement on `scene.name` instead of this, up to you?

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -109,6 +114,19 @@ protected virtual void OnValidate()
...
+            {
+                UserInterfaceManager.Initialize();
+            }
+            
+            if (scene.name == ADDITIVE_WEAPON_SCENE_NAME)
```
Gursaanj, line 124: can make this `else if` if you don't want to do a switch statement. Cause `scene.name` cant be both cases (useless check). 

Alternatively, you can do an early return after `UserInterfaceManager.Initialize()`

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -79,26 +83,27 @@ protected virtual void Awake()
...
-            SceneManager.LoadScene("AdditiveWeaponManager", LoadSceneMode.Additive);
-            SceneManager.LoadScene("AdditiveUserInterfaceScene", LoadSceneMode.Additive);
+            SceneManager.LoadScene(ADDITIVE_WEAPON_SCENE_NAME, LoadSceneMode.Additive);
+            SceneManager.LoadScene(ADDITIVE_USERINTERFACE_SCENE_NAME, LoadSceneMode.Additive);
 #if (UNITY_EDITOR || DEVELOPMENT_BUILD)
-            SceneManager.LoadScene("AdditiveDebugScene", LoadSceneMode.Additive);
+            SceneManager.LoadScene(ADDITIVE_DEBUG_SCENE_NAME, LoadSceneMode.Additive);
 #endif
+            
+            SceneManager.sceneLoaded += OnAdditiveSceneLoaded;
```
Gursaanj, line 92: Shouldn't this be added before the SceneManager.LoadScene calls? Just in case a scene gets loaded before this statement is reached (although realistically not possible I guess)

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -10,6 +10,10 @@ namespace BForBoss
 {
     public abstract class BaseWorldManager : MonoBehaviour
     {
+        private const string ADDITIVE_WEAPON_SCENE_NAME = "AdditiveWeaponManager";
+        private const string ADDITIVE_USERINTERFACE_SCENE_NAME = "AdditiveUserInterfaceScene";
```
Gursaanj, line 14: Super Nitpick : `ADDITIVE_USER_INTERFACE_SCENE_NAME`

Assets/_BForBoss/_Core/Scripts/UserInterface/GameplaySettingsViewBehaviour.cs
```diff
@@ -50,12 +50,10 @@ private void SetViews()
             _showRAMToggle.isOn = false;
         }
 
-        private void HideSettingsForProduction()
+        private void ShowSettingsForDevelopment()
```
Gursaanj, line 53: You can wrap this method with `#if DEVELOPMENT_BUILD || UNITY_EDITOR` since it will be useless in production anyway. 

Realistically the compiler will know that there is no reference to this and byte strip it away, but for the sake of consistency can just add the Pragma

Gursaanj (approved):

PR #278, opened 2022-09-22
BFB-383: Add Spawn and Death visuals

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-383

**Description**
- Applied spawning in VFX,
- Applied Death

**Screenshots / GIFs**
https://youtu.be/s_h9cxuFNes


Assets/_BForBoss/_VFX/Scripts/TimedVFXEffect.cs
```diff
@@ -32,7 +34,10 @@ public void StartEffect()
         private async void StopAfterTime()
```
Gursaanj, line 34: Just realized, I'm not sure why this needs to be async? (instead of coroutine). If this was just happening once, then it wouldn't make a difference, but `async` is quite heavy on garbage collection in comparison to coroutines ([reference](https://www.linkedin.com/pulse/unity-async-vs-coroutine-jo%C3%A3o-borks/#:~:text=While%20analyzing%20allocation,of%20operations%20simultaneously.))

And since this is being used on every enemy, it might be worth to switch?

Gursaanj (commented):
Looks awesome!

csweet22:
(Not a review, but this looks great! Thanks Martin! 🙌 )

Assets/_BForBoss/_VFX/Scripts/TimedVFXEffect.cs
m4rtint: Updated to be a coroutine

Gursaanj (approved):

PR #283, opened 2022-09-27
BFB-422: Player gets hurt when interacting with the Acid Ooze

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-422

**Description**
The player will now take damage when colliding with the Acid Ooze. the amount of damage as well as how much interval time between damage effects can be adjusted as needed

The TriggerAreaEffect is made so that we can also make areas heal the player as well if desired.

**Screenshots / GIFs**
TriggerArea


m4rtint:
That'd strange. There's no damage HUD effect?

Gursaanj:
> That'd strange. There's no damage HUD effect?

There is, it just really light! Probably because I compressed the GIF like a lot. If you look at the left hand corner when I first enter the water and then like every 3 seconds after

Assets/_BForBoss/_Core/Scripts/Environment/TriggerAreaEffect.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            if (_playerLifeCycle == null)
+            {
+                return;
+            }
+            
+            EffectPlayer();
+        }
+
+        private void Update()
```
m4rtint, line 44: Can you try to use onTriggerStay instead?

Assets/_BForBoss/_Core/Scripts/Environment/TriggerAreaEffect.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (_elapsedTime <= 0.0f)
+            {
+                EffectPlayer();
+                _elapsedTime = _secondsBetweenEffect;
+            }
+        }
+        
+        private void OnTriggerExit(Collider other)
+        {
+            if (_playerLifeCycle == null)
```
m4rtint, line 61: doesn't hurt to remove this if

Assets/_BForBoss/_Core/Scripts/Environment/TriggerAreaEffect.cs
```diff
@@ -0,0 +1,83 @@
...
+    [RequireComponent(typeof(Collider))]
+    public class TriggerAreaEffect : MonoBehaviour
+    {
+        private enum Effector
+        {
+            Damage,
+            Heal
+        }
+
+        [SerializeField] private Effector _effectOnPlayer = Effector.Damage;
```
m4rtint, line 15: This doesn't allow 2+ effectors on the player. E.g. Slow and Damage.

Gursaanj:
Added Modifiers so that the Player can be slowed/sped up upon entering the trigger area

TriggerArea Speed

Assets/_BForBoss/_Character/Scripts/PlayerMovementBehaviour.cs
```diff
@@ -68,6 +68,17 @@ public void SetControlConfiguration()
             characterLook.controllerHorizontalSensitivity = _inputConfiguration.ControllerHorizontalSensitivity;
             characterLook.controllerVerticalSensitivity = _inputConfiguration.ControllerVerticalSensitivity;
         }
+
+        public void ModifyPlayerSpeed(int modificationPercentage, out float originalPlayerSpeed)
+        {
+            originalPlayerSpeed = maxWalkSpeed;
```
m4rtint, line 74: You should keep an `_originalPlayerSpeed` within this script instead of outside. 

Having it outside means having 2 sources of truth of "original speed" and get confusing.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,127 @@
...
+
+namespace BForBoss
+{
+    [RequireComponent(typeof(Collider))]
+    public class PlayerTriggerAreaEffect : MonoBehaviour
+    {
+        private enum PlayerEffect
+        {
+            Damage,
+            Heal
```
m4rtint, line 15: Why do you have 2 separate ones. 
it'll be confusing to have 2 different enums. MIght as well put them all under PlayerEffect.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        private enum PlayerModifier
+        {
+            SlowDown,
+            SpeedUp
+        }
+
+        [Title("Player Effect", "Effect on Player over time", TitleAlignments.Split)]
+        [SerializeField] private PlayerEffect _playerEffect = PlayerEffect.Damage;
+        [SerializeField, Min(0), Tooltip("The amount to either damage or heal the player")] private int _effectAmount;
```
m4rtint, line 26: the name `_effectAmount` is not detailed enough at glance, is it health? speed?

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,127 @@
...
+        [SerializeField, Min(0.0f), Tooltip("Duration between player getting effected on")] private float _secondsBetweenEffect = 1.0f;
+
+        [Title("Player Modifier", "Modification on Player's stats", TitleAlignments.Split)]
+        [SerializeField] private PlayerModifier _playerModifier = PlayerModifier.SlowDown;
+        [SerializeField, Range(0, 100), Tooltip("Modification on Player stats as a percentage")]
+        private int _modificationAmount = 50;
```
m4rtint, line 32: It's not really clear what stats this is effecting. Is health? or speed? 
Also percentages are normally between 0 - 1

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
m4rtint: Okay I see the issue here. 
So the biggest issue is that `PlayerEffect` and `PlayerModifier` is basically the same meaning to others, and can get confusing. 

if u want to keep it as two enums, you can, but you have to be explicit to these meanings. like
`enum HealthChange` `enum SpeedChange` 

Or second thing, is to create a struct with the mutators. 
```
[System.Serializable]
public struct Mutator {
     enum PlayerEffect { Health, Speed }
    float percentage
}
```

And then you pass in this mutator in via inspector as an array of mutator. This is more future proof, but a bit more work.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
Gursaanj: Can definitely change it to 0 - 1

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
Gursaanj: Went ahead and made it a class instead so that I can add default values to certain fields!

Gursaanj:
This is what the inspector looks like atm. Ensuring that designers will know which are effects (the thing that affects the player every x second) and which are modifiers (the things that affects the player only once immediately upon entering)
inspecotr

Assets/_BForBoss/_Character/Scripts/PlayerMovementBehaviour.cs
```diff
@@ -69,6 +73,17 @@ public void SetControlConfiguration()
             characterLook.controllerVerticalSensitivity = _inputConfiguration.ControllerVerticalSensitivity;
         }
         
+        public void ModifyPlayerSpeed(float modificationMultiplier)
+        {
+            _unModifiedPlayerSpeed = maxWalkSpeed;
+            maxWalkSpeed += maxWalkSpeed * modificationMultiplier;
```
m4rtint, line 79: Does `maxWalkSpeed *= modificationMultiplier`

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        private enum SpeedModifier
+        {
+            SlowDown,
+            SpeedUp
+        }
+        
+        
+        [Serializable]
+        private class PlayerMutation
```
m4rtint, line 27: Can this be placed in a separate file, And also don't call it `PlayerMutation` 
Just call it `Mutation`. 
PlayerMutation implies it's only for players, as opposed to anything else.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,142 @@
...
+            _playerLifeCycle = other.GetComponent<PlayerLifeCycleBehaviour>();
+            _playerMovementBehaviour = other.GetComponent<PlayerMovementBehaviour>();
+
+            if (_playerLifeCycle == null || _playerMovementBehaviour == null)
+            {
+                return;
+            }
+            
+            EffectPlayer();
+            ModifyPlayerStats();
```
m4rtint, line 75: EffectPlayer();
ModifyPlayerStats();

Are essentialy the same thing, so you should only have one method that effects the player. Since it's confusing as to why there's two of the same methods being called.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
```diff
@@ -0,0 +1,142 @@
...
+    [RequireComponent(typeof(Collider))]
+    public class PlayerTriggerAreaEffect : MonoBehaviour
+    {
+        private enum HealthEffect
+        {
+            Damage,
+            Heal
+        }
+
+        private enum SpeedModifier
```
m4rtint, line 19: I don't think you need these enum's anymore ,since you have the `Serializable` PlayerMutation

Assets/_BForBoss/_Character/Scripts/PlayerMovementBehaviour.cs
Gursaanj: This could work with `SlowDown` but with `SpeedUp` then the multiplier of 0 makes the player motionless, and the multiplier of 1 will make the player move at its normal speed.

So I think by doing `+=` keeps it more intuitive

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
Gursaanj: I think `Mutation` is to vague as it can refer to anything that mutates, maybe too be more specific `PlayerTriggerAreaMutation`??

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
Gursaanj: Effect - called repeatedly every `_secondsBetweenEffect` seconds.
Modification - called once when you enter in the trigger zone - and reset when you leave

It is way more intuitive to keep those two different concepts separate!

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
Gursaanj: Do you mean to place these enums within the `PlayerMutation` class?

m4rtint: No, to literally just remove them.

You can just setup the mutation struct object to have the correct settings.

Having this complicates things with
2 different enums + struct to wrap your head around, and gets confusing

m4rtint:
> This is what the inspector looks like atm. Ensuring that designers will know which are effects (the thing that affects the player every x second) and which are modifiers (the things that affects the player only once immediately upon entering)
> inspecotr
> 

So relative to removing the enum

Im thinking its just easier to remove the enum and allow users to set the properties as
(-)10
Like they can manually set it. Don't need an enum.

Assets/_BForBoss/_Core/Scripts/Environment/PlayerTriggerAreaEffect.cs
m4rtint: In that case can you update the naming of the method to make it clear. One time effect vs repeated effect. And use the same jargon, either use effect or status change. 

Stick with one. Using Different ones implies different meanings

m4rtint (approved):

PR #293, opened 2022-10-17
BFB-454: Health Pick up Full Health

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-455

**Description**
Checks if player is at full health before picking up item

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/196077515-2ffc5644-18ad-4dbd-b51e-cab44935032f.mp4

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Environment/ItemPickupBehaviour.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Environment/ItemPickupBehaviour.cs
@@ -27,7 +27,7 @@ namespace BForBoss
         
         private void OnTriggerEnter(Collider other)
         {
-            if (other.TryGetComponent(out PlayerLifeCycleBehaviour _))
+            if (other.TryGetComponent(out PlayerLifeCycleBehaviour playerLife) && !playerLife.IsFullHealth)
             {
                 IsShown(false);
             }
--- a/Assets/_BForBoss/_Core/Scripts/Environment/HealingBehaviour.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Environment/HealingBehaviour.cs
@@ -8,7 +8,7 @@ namespace Perigon.Entities
         
         private void OnTriggerEnter(Collider other)
         {
-            if (other.TryGetComponent(out PlayerLifeCycleBehaviour lifeCycle))
+            if (other.TryGetComponent(out PlayerLifeCycleBehaviour lifeCycle) && !lifeCycle.IsFullHealth)
             {
                 lifeCycle.Heal(_healAmount);
             }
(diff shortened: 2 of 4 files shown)


Assets/_BForBoss/_Core/Scripts/Environment/ItemPickupBehaviour.cs
```diff
@@ -27,7 +27,7 @@ public void CleanUp()
         
         private void OnTriggerEnter(Collider other)
         {
-            if (other.TryGetComponent(out PlayerLifeCycleBehaviour _))
+            if (other.TryGetComponent(out PlayerLifeCycleBehaviour playerLife) && !playerLife.IsFullHealth)
```
Gursaanj, line 30: This would mean all items (not just items that heal you supposedly) would not work if Player Health is 100% right?

Not sure that seems like the correct behaviour.

Gursaanj: An idea would be to have `HealingBehaviour` derive from something like `ItemBehaviour` in which it has a virtual method i.e. `public virtual bool CanPickUp()`.

and instead `_itemToPickUp` being a `GameObject` it would be a `ItemBehaviour`. Thus, this check would be 

`if (other.TryGetComponent(out PlayerLifeCycleBehaviour playerLife) && _itemToPickup.CanPickUp())`

What do you think?

m4rtint: >This would mean all items (not just items that heal you supposedly) would not work if Player Health is 100% right?

that's the point of the ticket, that it won't work when it's 100% or more.

If it's another type of pickup item, it won't enter the if statement and just go to the next statement. E.g.

```
if (TryGet && IsFullHealth) {
    doSomething();
}

if (TryGet && SomethingElse) {
    doSomethingElse();
}
```

m4rtint: >Not sure that seems like the correct behaviour.

You can Try it out in game and see if u can find any edge cases that won't work? 
Basically if it's <100%, it will work.

m4rtint:
@Gursaanj restructured the pick up health item

Assets/_BForBoss/_Core/Scripts/Environment/HealingItemBehaviour.cs
```diff
@@ -0,0 +1,23 @@
+using BForBoss;
+using UnityEngine;
+
+namespace Perigon.Entities
+{
+    public class HealingItemBehaviour : ItemPickupBehaviour
+    {
+        [SerializeField] private float _healAmount = 50f;
+
+        protected override bool DidPickedUpItem(Collider other)
```
Gursaanj, line 10: Grammatically incorrect. Either `DidPickUpItem` or `HasPickedUpItem` would work

Gursaanj (approved):
Nitpick change only

PR #297, opened 2022-10-20
BFB-466: Resolve OffMeshLinker null Reference

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-466

**Description**
In certain cases where the Agent has to pop into the navmesh offMeshLink (in this case, due to a ledge) the agent's `OffMeshLink` might be null even if the associated `CurrentOffMeshLinkData` is valid. So a null check was added to ensure no null references

**Screenshots / GIFs**
offmeshLinkJumpIssue


Assets/_BForBoss/_Entities/Enemies/FlyingRobot/Scripts/OffMeshLinkChecker.cs
```diff
@@ -30,8 +30,14 @@ private void AcquireOffMeshLink()
             {
                 return;
             }
-            
+
             _link = _agent.currentOffMeshLinkData.offMeshLink;
+
+            if (_link == null)
```
m4rtint, line 36: this whole method is a bit hard to understand.

m4rtint (approved):

PR #298, opened 2022-10-21
BFB-445: No overlapping enemy spawn

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-445

**Description**
- Check if there's an overlap in position
- Retries 10 times max before leaving it alone. 

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/197094464-ba2f930e-5962-42ae-be09-f82dd9a6568f.mp4


Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
```diff
@@ -1,8 +1,13 @@
 using System.Collections;
 using Perigon.Entities;
+using Perigon.Utility;
+using Sirenix.Utilities;
 using UnityEngine;
 using UnityEngine.AI;
 using Logger = Perigon.Utility.Logger;
+using Quaternion = UnityEngine.Quaternion;
+using Random = UnityEngine.Random;
+using Vector3 = UnityEngine.Vector3;
```
Gursaanj, line 10: is there another `Vector3` being used? Not sure if this is needed?

Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
```diff
@@ -86,18 +93,53 @@ private IEnumerator SpawnEnemies(int count)
...
+                    return hit.position;
+                }
+            }
+
+            return null;
+        }
+
+        private bool IsPositionValid(Vector3 position)
+        {
+            const int halfExtentSizePositionValidityCheck = 2;
```
Gursaanj, line 140: what is supposed to do exactly?

Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
```diff
@@ -12,6 +17,8 @@ public class EnemySpawnAreaBehaviour : MonoBehaviour
         [SerializeField] [Min(0.0f)] private float _secondsBetweenSpawns = 2.5f;
         [SerializeField, Tooltip("Should max amount of enemies spawn upon initialization")]
         private bool _burstInitialSpawn = true;
+        [SerializeField] 
+        private float _spawnArea = 1;
```
Gursaanj, line 21: What is this supposed to be? A multiplier for the positioning?

Does this have to be exposed? If so, a better naming needs to be used

m4rtint: yeah i just updated it. However i'm thinking of a better way to phrase this since it's a bit of a lie. 

1. it first get's a random spot within a radius
2. It then get's closest navmesh point. HOWEVER, the closest navmesh point might not be within that specific radius. 

Sooooooooo, i'm trying to think of something a bit more intuitive, or just not expose it at all.

Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
m4rtint: `Physics.OverlapBox()`
https://docs.unity3d.com/ScriptReference/Physics.OverlapBox.html

half extents, for the size of the overlap box

Gursaanj: fair. personally, I think it would more readable to just do `Vector3 halfExtents = Vector3.one * 2;` but I leave that up to you!

Gursaanj: To be honest, I would like the actual half extent to be the half extent of the enemy model so that if we change its dimensions or create another enemy to use with completely different proportions, this would be adjusted accordingly. We could technically get that info since we have access to the enemy from `EnemyContainer`, but it could be a bit tedious. What do you think?

m4rtint: It's kinda overkill for this.

Gursaanj: For this task I'm down to keep this as is, but I will probably make a backlog task to make the appropriate change

Assets/_BForBoss/_Core/Scripts/Environment/EnemySpawnAreaBehaviour.cs
```diff
@@ -12,6 +15,8 @@ public class EnemySpawnAreaBehaviour : MonoBehaviour
         [SerializeField] [Min(0.0f)] private float _secondsBetweenSpawns = 2.5f;
         [SerializeField, Tooltip("Should max amount of enemies spawn upon initialization")]
         private bool _burstInitialSpawn = true;
+        [SerializeField] 
+        private float _spawnRadius = 1;
```
Gursaanj, line 19: maybe `_spawnRadiusMultiplier`? and to ensure no one puts a crazy value, you can define a range (i.e. between 0 and 3 maybe??) - at least add a Min attribute to ensure no negative values are used.

Also think a Tooltip would be really useful

m4rtint: It's not a multiplier, it's literally just the size of the box. it's not multiplying anything.

Gursaanj (approved):

PR #299, opened 2022-10-21
BFB-458: Respawn Healing Items at the end of wave if desired

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-458

**Description**
If ticked true via the world manager, after being consumed, healing items will only respawn at the start of the next wave


Assets/_BForBoss/_Core/Scripts/Environment/HealingItemBehaviour.cs
```diff
@@ -6,6 +6,14 @@ namespace Perigon.Entities
     public class HealingItemBehaviour : ItemPickupBehaviour
     {
         [SerializeField] private float _healAmount = 50f;
+        private bool _shouldResetOnlyWhenWaveIncrements = false;
+        private int _currentWaveNumber;
```
m4rtint, line 10: The issue with putting this logic into this Item, is that this item is now explicitly tied to the context of waves. 

The concept of waves shouldn't be known by the items. But The waveWorldManager can know about it.  If we were to move this to another game. E.g. Mario with healing items, this wouldn't make sense.

Assets/_BForBoss/_Core/Scripts/Environment/HealingItemBehaviour.cs
```diff
@@ -6,6 +6,14 @@ namespace Perigon.Entities
     public class HealingItemBehaviour : ItemPickupBehaviour
     {
         [SerializeField] private float _healAmount = 50f;
+        private bool _shouldResetOnlyWhenWaveIncrements = false;
```
m4rtint, line 9: Don't think you'll need an option to switch this. Since it's a global change, designers will probably touch this at most once or twice, and then never touch it again. 

Should probably setup the code in a way where you can revert back to timer style easily.

Assets/_BForBoss/_Core/Scripts/Environment/HealingItemBehaviour.cs
Gursaanj: So do you think like a `ItemTimer` Monobehaviour Component that we attach as a dependancy on `HealingItemBehavioiur` (i.e. [SerializeField] private ItemTimer _itemTimer) and then call on its method when needed?

Something like

`protected override void Update() {_itemTimer.OnUpdate}`? Then at least the item has no direct dependency on the Wave system.

m4rtint: Yeah, i think you would need to separate the timing aspect into an `ItemRespawnTimer` class. 
`ItemTimer` is too vague as to what you're referring to.

However, you wouldn't need to call `_itemTimer.onUpdate` and let it be handled automatically, since it's not really related to the pick up actions/logic.

Gursaanj:
Created two different Item Respawn behaviours, one to respawn the item based on time (i.e. the way we have implemented it now) and one to respawn the item after the wave has ended. 

With my PR I made our Healing Item to use the wave based respawn behaviour.

m4rtint (approved):

PR #302, opened 2022-10-31
BFB-473 : Stop Player from being shot while End Game Panel is open

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-473

**Description**
Time is now paused while in the death state (i.e. when the end game behaviour is open)

### DIFF
--- a/Assets/_BForBoss/_Utility/Scripts/CustomPassVolumeWeightTool.cs
+++ b/Assets/_BForBoss/_Utility/Scripts/CustomPassVolumeWeightTool.cs
@@ -37,12 +37,12 @@ namespace Perigon.Utility
         
         public Tweener Distort()
         {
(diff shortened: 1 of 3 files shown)


Assets/_BForBoss/_Utility/Scripts/CustomPassVolumeWeightTool.cs
```diff
@@ -37,12 +37,12 @@ public CustomPassVolumeWeightTool(CustomPassVolume customPassVolume, string mate
         
         public Tweener Distort()
         {
-            return DOTween.To(intensity => _material.SetFloat(_materialKey, intensity), _startValue, _endValue, _distortDuration);
```
Gursaanj, line 40: This ensures that the tweens use unscaled time. Otherwise, the blackscreen never shows up and the tween never completes

m4rtint (approved):

PR #306, opened 2022-11-04
Audio Integration

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-282
https://perigongames.atlassian.net/browse/BFB-290
https://perigongames.atlassian.net/browse/BFB-462
https://perigongames.atlassian.net/browse/BFB-419

**Description**
Implements all new audio, tweaks volumes across the board (including lowering the footstep audio) and confirms that audio settings are adjusted correctly now. 

- Includes audio for AI movement, shooting, charging, and updated death effect (on hit effect is unchanged)
- Includes new player audio for footsteps, as well as healing, melee hits and misses. 
- - Walking audio is parameterized with player speed, note that this is only noticeable with controller since MKB immediately goes to full speed
- Includes reload audio for guns, and new firing audio
- - AR audio changes as fire rate increases
- Adds new, quieter BGM


Assets/_BForBoss/_Character/Scripts/CharacterMovementAudio.cs
```diff
@@ -42,6 +42,13 @@ public void OnUpdate()
                 OnSoundStateChange?.Invoke(_state);
             }
         }
+
+        public float GetPlayerSpeedNormalized()
+        {
+            return _state == MovementSoundState.Running
```
Gursaanj, line 48: I know this might be overkill but I think for safety, we should check that `_characterMovement.CharacterMaxSpeed` != 0. so that we don't get a divide by 0 error!

Assets/_BForBoss/_Entities/Scripts/LifeCycleBehaviour.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
Gursaanj, line 1: wait is this needed?

Assets/_BForBoss/_Character/Scripts/CharacterMovementAudio.cs
Gursaanj: i.e. `Math.max(_characterMovement.CharacterMaxSpeed, 1)`

Assets/_BForBoss/_Weapons/Scripts/Weapons/AutomaticWeaponBehaviour.cs
```diff
@@ -1,17 +1,32 @@
...
+        }
+        
         protected override void OnFireInputAction(bool isFiring)
         {
             _isFiring = isFiring;
             
             if (!_isFiring)
             {
                 _timeSinceFire = 0;
+                _shotsFired = 0;
```
Gursaanj, line 29: So `_timeSinceFire` is set to 0 when `WeaponBehaviour.Reset()` is [called](https://github.com/PerigonGames/BForBoss/blob/ca58f30fe50539115d5fdf7c48e001cb0f13b1ed/Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponBehaviour.cs#L66), should we make that method virtual and override it here with `_shotsFired = 0`?

m4rtint (approved):
Some of the weapon audio I picked out definitely needs some changes. But all in all, good placements

Assets/_BForBoss/_Weapons/Scripts/Weapons/MeleeWeapon.cs
```diff
@@ -30,26 +30,18 @@ public void DecrementCooldown(float deltaTime)
...
-                return false;
-            _currentCooldown += _meleeProperties.AttackCoolDown;
-            
-            _hits = _meleeProperties.OverlapCapsule(playerPosition, playerForwardDirection, TagsAndLayers.Layers.PlayerMask, ref _enemyBuffer);
-            if (_hits > 1)
-                _hits = 1; //ensure we only damage first enemy
-            return true;
+            var hits = TryAttack(playerPosition, playerForwardDirection);
+
+            if (!hits.HasValue) return null;
```
Gursaanj, line 42: Instead of making this nullable, could we instead return -1 to signify a a failed `TryAttack()`? 

Personal preference, just find nullables a bit harder to read and like to avoid them.

Assets/_BForBoss/_Weapons/Scripts/Weapons/MeleeWeapon.cs
```diff
@@ -66,6 +58,16 @@ public IList<Vector3> ApplyDamage(Vector3 position)
             return pointsHit;
         }
 
+        private int? TryAttack(Vector3 playerPosition, Vector3 playerForwardDirection)
+        {
+            if (!CanMelee)
+                return null;
```
Gursaanj, line 64: same here with the -1 return instead of null?

Assets/_BForBoss/_Weapons/Scripts/Weapons/Weapon.cs
```diff
@@ -16,8 +16,19 @@ public class Weapon
...
 
-         public bool IsReloading { get; set; }
+         public bool IsReloading
+         {
+             get => _isReloading;
+             set
+             {
+                 if (value == _isReloading) return;
+                 _isReloading = value;
+                 if (_isReloading) OnStartReloading?.Invoke();
```
Gursaanj, line 28: could you put these in curly braces

Assets/_BForBoss/_Weapons/Scripts/Weapons/Weapon.cs
```diff
@@ -108,17 +121,18 @@ public Vector3 GetShootDirection(float timeSinceFiring)
              return new Vector3(xPosition, yPosition, 1);
          }
 
-         public void FireIfPossible()
+         public bool FireIfPossible()
```
Gursaanj, line 124: maybe a better method name would be `TryFire()` since its returning a bool. What do you think?

Assets/_BForBoss/_Weapons/Scripts/Weapons/WeaponBehaviour.cs
```diff
@@ -81,12 +81,18 @@ private void HandleOnStopReloading()
         {
             _weaponAnimationProvider.ReloadingWeapon(false);
         }
+        
+        protected virtual void HandleOnStartReloading()
```
Gursaanj, line 85: protected should be above private

Gursaanj: realized this whole class has a mishmash of protected and private, will leave it up to you if you want to change things around

Gursaanj (commented):
Mostly small comments

Assets/_BForBoss/_Weapons/Scripts/Weapons/Weapon.cs
unoctium1: TryFire makes sense to me, I'll rename it

Assets/_BForBoss/_Weapons/Scripts/Weapons/AutomaticWeaponBehaviour.cs
unoctium1: Ah good catch, I'll do that!

Assets/_BForBoss/_Entities/Scripts/LifeCycleBehaviour.cs
unoctium1: No good catch, I'd originally made changes here and forgot to remove the usings when I undid them

PR #315, opened 2023-02-13
BFB-487: Individual Wall Run

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-487

**Description**
- Created WallRunSO (ScriptableObject)
- Create WallRunData - Pure C#
- Create WallRunContainer - For the Walls
- PlayerWallRun get's the data.

**Screenshots / GIFs**
image
image


Assets/_BForBoss/_Core/Scripts/Environment/WallRunning/WallRunSO.cs
```diff
@@ -0,0 +1,66 @@
...
+        [SerializeField]
+        private float _cameraRotateDuration = 1f;
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField] 
+        private float _lookAlongWallRotationSpeed = 3f;
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField] 
+        private float _minLookAlongWallStabilizationAngle = 5f;
+
+        public WallRunData MapToData()
```
m4rtint, line 49: I'm going to start refactoring scriptable objects in the future. 

Realized we shouldn't be making them implement interfaces, Should just have a "Domain" object - pure data, and a Scriptable object - Designer facing for value changes. 

So for testing/tools, you just need to create a pure data "domain" object and pass it into whatever needs it. Act's the same way as an interface, i just used an interface incorrectly.

Assets/_BForBoss/_Core/Scripts/Character/PlayerWallRunBehaviour.cs
```diff
@@ -404,9 +369,9 @@ private float GetCameraRoll()
             float targetAngle = 0;
             if (wallDirection != 0)
             {
-                targetAngle = Mathf.Sign(wallDirection) * heading * _maxCameraAngleRoll;
+                targetAngle = Mathf.Sign(wallDirection) * heading * _wallRunData.MaxCameraAngleRoll;
             }
-            return Mathf.LerpAngle(cameraAngle, targetAngle, Mathf.Max(_timeSinceWallAttach, _timeSinceWallDetach) / _cameraRotateDuration);
+            return Mathf.LerpAngle(cameraAngle, targetAngle, Mathf.Max(_timeSinceWallAttach, _timeSinceWallDetach) / _wallRunData.CameraRotateDuration);
```
Gursaanj, line 374: For safety reasons, we should check to see if `_wallRunData.CameraRotateDuration` is 0 first (and if that is the case then divide by 1). 

Alternatively, we can add a min value for `CameraRotateDuration`. Would still prefer if we do the check as well.

Assets/_BForBoss/_Core/Scripts/Environment/WallRunning/WallRunSO.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField]
+        private float _maxCameraAngleRoll = 30f;
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField]
+        private float _cameraRotateDuration = 1f;
+        [FoldoutGroup("Camera Settings")]
+        [SerializeField] 
+        private float _lookAlongWallRotationSpeed = 3f;
```
Gursaanj, line 44: I think this field and `_minLookAlongWallStabilizationAngle` should have tooltips as well

Assets/_BForBoss/_Core/Scripts/Character/PlayerWallRunBehaviour.cs
m4rtint: So default values in struct can only be the built in C# default value which is 0. 

So i'm setting up a Max(cameraiValue, 0.01)

Assets/_BForBoss/_Core/Scripts/Environment/WallRunning/WallRunSO.cs
m4rtint: Updated

Gursaanj (approved):

PR #320, opened 2023-02-26
BFB-493: Initial Data/ScriptableObject files for Energy System

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-493

**Description**
Created the initial data files for the Energy System and its corresponding Configuration

The data is meant to hold information about how much energy each Transaction holds and the rate at which it transfers to/from the player

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
+++ b/Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
@@ -1,0 +1,25 @@
+using UnityEngine;
+
+namespace BForBoss
+{
+    [CreateAssetMenu(fileName = "Energy Data", menuName = "PerigonGames/EnergySystem/EnergyData", order = 1)]
+    public class EnergySO : ScriptableObject
+    {
+        [SerializeField, Tooltip("Player's current value of Energy"), Min(0f)] 
+        private float _value = 0f;
+
+        [SerializeField, Tooltip("Max Energy Attainable")]
+        private float _maxEnergyValue = 100f;
+
+        [SerializeField, Tooltip("How often does the player gain energy per second when in the middle of accruement")]
+        private float _rateOfAccruement = 1f;
+
+        public EnergyData MapToData()
+        {
+            return new EnergyData(
+                value:_value,
+                maxEnergyValue: _maxEnergyValue,
+                rateOfAccruement: _rateOfAccruement);
+        }
+    }
+}
--- a/Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
+++ b/Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
@@ -5,21 +5,21 @@ namespace BForBoss
     [CreateAssetMenu(fileName = "Energy Data", menuName = "PerigonGames/EnergySystem/EnergyData", order = 1)]
     public class EnergySO : ScriptableObject
     {
-        [SerializeField, Tooltip("Player's current value of Energy"), Min(0f)] 
-        private float _value = 0f;
+        [SerializeField, Tooltip("Player's starting value of Energy"), Min(0f)] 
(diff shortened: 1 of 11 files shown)


Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
```diff
@@ -0,0 +1,25 @@
+using UnityEngine;
+
+namespace BForBoss
+{
+    [CreateAssetMenu(fileName = "Energy Data", menuName = "PerigonGames/EnergySystem/EnergyData", order = 1)]
+    public class EnergySO : ScriptableObject
+    {
+        [SerializeField, Tooltip("Player's current value of Energy"), Min(0f)]
```
Gursaanj, line 8: Not sure if these tooltips are accurate, let me know if I should edit them

Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class EnergySO : ScriptableObject
+    {
+        [SerializeField, Tooltip("Player's current value of Energy"), Min(0f)] 
+        private float _value = 0f;
+
+        [SerializeField, Tooltip("Max Energy Attainable")]
+        private float _maxEnergyValue = 100f;
+
+        [SerializeField, Tooltip("How often does the player gain energy per second when in the middle of accruement")]
+        private float _rateOfAccruement = 1f;
```
Gursaanj, line 15: Should we also have a `_rateOfExpense` field since things like SlowMo is an expense that also goes at a rate

BForBoss.sln.DotSettings
```diff
@@ -1,2 +1,4 @@
 ﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Accruement/@EntryIndexedValue">True</s:Boolean>
```
Gursaanj, line 2: Accruement is a word ([link](https://www.vocabulary.com/dictionary/accruement)) but Rider doesn't seen to know that, so added it to our solution's dictionary

Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
m4rtint: Hmmm I think this rate should be for both Accurument and Expense. 
To keep it consistent between both, and not expose too many dynamic variable (keep it simple). 

So this should be called `_rateOfChange` or if you have a better name that encomposses either accrument or expense

Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
m4rtint: Yes that is accurate. hmm, I think it's better to call this `_startingValue`. 

But leave the `Energy.Value` as just value.

BForBoss.sln.DotSettings
m4rtint, line 2: Can you revert this file?

m4rtint:
Everything looks good, just need to change some variable names, and it's good to go

Assets/_BForBoss/_Core/Scripts/EnergySystem/Data/EnergySO.cs
Gursaanj: Will call it `_rateOfTransaction` to keep in line with our documentation I guess!

m4rtint (approved):

PR #321, opened 2023-02-27
BFB-502: Add Color Param for the logger tool

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-502

**Description**
Added a color parameter to our Logger Tool methods, so that people can more easily set colors to their logged messages

**Screenshots / GIFs**
grab 1
grab 2

### DIFF
--- a/Assets/_BForBoss/_Utility/Scripts/Logger/Logger.cs
+++ b/Assets/_BForBoss/_Utility/Scripts/Logger/Logger.cs
@@ -2,14 +2,23 @@ using UnityEngine;
 
 namespace Perigon.Utility
 {
+    public enum LoggerColor
+    {
+        Black,
+        Red,
+        Blue,
+        Green,
+        Yellow
+    }
+    
     public static partial class Logger
     {
-        public static void LogString(string toLog, string key = "Misc")
+        public static void LogString(string toLog, LoggerColor color = LoggerColor.Black,  string key = "Misc")
         {
 #if !UNITY_EDITOR && DEBUG
-            Debug.Log(toLog);
+            Debug.Log(ColorizeLog(toLog, color));
 #elif UNITY_EDITOR
-            LogInEditor(toLog, key, Debug.Log);
+            LogInEditor(ColorizeLog(toLog, color), key, Debug.Log);
 #endif
         }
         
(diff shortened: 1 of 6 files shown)


Assets/_BForBoss/_Utility/Scripts/Logger/Logger.cs
```diff
@@ -2,14 +2,23 @@
...
+        public static void LogString(string toLog, LoggerColor color = LoggerColor.Black,  string key = "Misc")
         {
 #if !UNITY_EDITOR && DEBUG
-            Debug.Log(toLog);
+            Debug.Log(ColorizeLog(toLog, color));
 #elif UNITY_EDITOR
-            LogInEditor(toLog, key, Debug.Log);
+            LogInEditor(ColorizeLog(toLog, color), key, Debug.Log);
 #endif
         }
```
Gursaanj, line 24: Decided to not add the parameter to our LogFormat method since that method will already be doing intentional special formatting to its text, so it seemed unnecessary to add

Assets/_BForBoss/_Utility/Scripts/Logger/Logger.cs
```diff
@@ -22,22 +31,51 @@ public static void LogFormat(string toLog, string key = "Misc", params object[]
...
         
-        public static void LogWarning(string toLog, string key = "Misc")
+        public static void LogWarning(string toLog, LoggerColor color = LoggerColor.Black, string key = "Misc")
         {
 #if !UNITY_EDITOR && DEBUG
-            Debug.LogWarning(toLog);
+            Debug.LogWarning(ColorizeLog(toLog, color));
```
m4rtint, line 37: Within the context of this ticket. this is a good solution.
 
However, to take it a step further, I would separate `ColorizeLog` into its own variable, and then pass it into `LogWarning`. 

This way, if we want more customized Logs (like the `toLogs` would already have its own tags from outside), it'll be easier to update it.

m4rtint (approved):
This lgtm, just a comment for future reference
https://github.com/PerigonGames/BForBoss/pull/321/files#r1120847564

PR #323, opened 2023-03-02
BFB-495: Player movement integrated with energy system

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-495

**Description**
Energy system within PlayerMovement
Removed Double Jump from accrument
Changed Dash to accrue energy on stop.

Added energy system to sandboxWorldManager

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/222331294-cd74a96e-ed6b-4bbc-8c4a-cf5d3a2e4715.mp4


Assets/_BForBoss/_Core/Scripts/Managers/SandboxWorldManager.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class SandboxWorldManager : BaseWorldManager
+    {
+        [SerializeField] private EnergySystemBehaviour _energySystemBehaviour;
+
+        protected override Vector3 SpawnLocation => _spawnLocation.position;
+        protected override Quaternion SpawnLookDirection => _spawnLocation.rotation;
+
+        protected override void Start()
+        {
+            _playerBehaviour.EnergySystem = _energySystemBehaviour;
```
m4rtint, line 14: This is something I'm not quite sure about. 
Should this be in the BaseWorldManager? 
if it's in the BaseWorldManager, Then the energy system would have to be added into each and every scene.

Assets/_BForBoss/_Core/Scripts/Character/PlayerDashBehaviour.cs
```diff
@@ -8,7 +8,7 @@ namespace BForBoss
 {
     public interface IPlayerDashEvents
     {
-        void OnDashStarted();
+        void OnDashStopped();
```
m4rtint, line 11: So I realized that `OnDash` get's executed few timers per frame. 

Which means accrue would get called more than once. 
However, OnDashStopped is guarenteed to be called once. 

So I set the accruement `onDashStopped` instead.

Assets/_BForBoss/_Core/Scripts/EnergySystem/EnergySystemBehaviour.cs
```diff
@@ -84,7 +93,11 @@ private void Start()
 
         private void Update()
         {
-            _elapsedTime -= Time.deltaTime;
+            if (_accruedEnergyTypeQueue.Count > 0 || _expendEnergyTypeQueue.Count > 0)
```
m4rtint, line 96: `Update` Get's called various times per frame. so the accruement vs disposing extra values get's super unbalanced and off. 

 So i'm changing it where it only counts down when there's something in the queue.

Assets/_BForBoss/_Core/Scripts/Managers/SandboxWorldManager.cs
Gursaanj: hmm good question, we can do this for now and put in BaseWorldManager in the future if we think that is better

Assets/_BForBoss/_Core/Scripts/PlayerBehaviour.cs
```diff
@@ -13,12 +13,13 @@ public class PlayerBehaviour : MonoBehaviour
         private PlayerLifeCycleBehaviour _playerLifeCycleBehaviour;
         private PlayerSlowMotionBehaviour _playerSlowMotion;
         private PGInputSystem _inputSystem;
+        public IEnergySystem EnergySystem;
```
Gursaanj, line 16: public above private

Gursaanj (approved):
One Small Nitpick, but looks good

PR #325, opened 2023-03-05
BFB-496: Slow motion with energy

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-496

**Description**
- Expend energy for Slow motion
- 
**Screenshots / GIFs**
https://user-images.githubusercontent.com/5779541/222992359-7dda2c04-81fe-4b49-9c9d-601ffbbc4b46.mp4


Assets/_BForBoss/_Core/Scripts/EnergySystem/EnergySystemBehaviour.cs
```diff
@@ -11,6 +11,7 @@ public interface IEnergySystem
     {
         void Accrue(EnergyAccruementType accruementType, float multiplier = 1);
         void Expend(EnergyExpenseType expenseType, float multiplier = 1);
+        bool CanExpend(EnergyExpenseType expenseType, float multiplier = 1);
```
Gursaanj, line 14: could make this `bool TryExpend()`? then we don't have to do `if CanExpend() {Expend()}` all the time

m4rtint: In this case, I needed to have 2 seperate methods. 

Check if you can activate.. 
then activate + all the other stuff that goes along + tweening + sound... 

then you have the other state - While it's activated (while it's in a loop).  Then check again if you can still expend, and then expend energy.

m4rtint: I think especially in this more generic interface, I rather have every method be clear cut, concise, and does only ONE simple function. 

And then in the concrete classes, u can do "Try" if need be. (use interface as building blocks, as opposed to force the concrete class to implement multiple functionality based on method name).

Gursaanj (approved):

PR #335, opened 2023-03-15
BFB-238: Dynamic crosshair

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-238

**Description**
- Created a dynamic crosshair. 
- Removed storing the crosshair in the weaponSO. 

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/225179721-9f436829-b0a4-4ec6-8e57-838fb4600f9d.mp4


unoctium1:
Just to understand this a bit better, is the idea that the crosshairs expand regardless of what gets hit?

m4rtint:
> Just to understand this a bit better, is the idea that the crosshairs expand regardless of what gets hit?

Yes

unoctium1 (approved):
Cool cool, in that case it looks great to me

PR #336, opened 2023-03-15
Create labor system and basic tests

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-499

**Description**
Create the initial (non monobehaviour) labor system, as well as the ILabor interface. Also wrote basic tests to demonstrate usage


Assets/_BForBoss/_Core/Scripts/Labor/LaborSystem.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public class LaborSystem
+    {
+        public Action onLaborCompleted;
+        public Action onQueueCompleted;
+        
+        private Queue<ILabor> _laborsToComplete;
+        public ILabor CurrentLabor { get; private set; }
+
+        public void Initialize(IEnumerable<ILabor> labors)
```
m4rtint, line 14: If we're not using Mono, this could be an actual constructor

Assets/_BForBoss/_Core/Scripts/Labor/LaborSystem.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+
+namespace BForBoss.Labor
+{
+    public class LaborSystem
+    {
+        public Action onLaborCompleted;
+        public Action onQueueCompleted;
```
m4rtint, line 9: Since this hasn't been decided yet, we don't need these Actions just yet. could just delete them for now.

Assets/_BForBoss/_Core/Scripts/Labor/LaborSystem.cs
unoctium1: Yeah I wasn't sure if we wanted to couple the labor system to a specific set of labors. I thought this seemed more reusable, ie one labor system that gets initialized multiple times with different sets of labors. I might be overcomplicating it though

unoctium1: Yeah I wasn't sure if we wanted to couple the labor system to a specific set of labors. I thought this seemed more reusable, ie one labor system that gets initialized multiple times with different sets of labors. I might be overcomplicating it though

Assets/_BForBoss/_Core/Scripts/Labor/LaborSystem.cs
unoctium1: Sure, I'm using the queue one to facilitate tests but I could make an exposed IsComplete boolean flag or something to test against instead

m4rtint: An idea i had, was to have an `event Action`, where it returns an enum that signifies an effect.
so something like
```
enum LaborSystemEffect {
   case LaborCompleted
   case QueueComplete
}

event Action<LaborSystemEffect> LaborSystemEffect
```

but this is for later. you can also check that `CurrentLaborSystem == null`

unoctium1: That makes sense yeah. The more I think about too though I think having an IsComplete property seems useful

m4rtint (approved):

PR #340, opened 2023-03-19
Potential Solution

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-[TICKET NUMBER]

**Description**
A solution to the bug where Player hits themselves when spawning a bullet due to the BulletSpawner is, Create two different sets of pools in which each set of pools generates bullets with Layermasks that are appropriate with Players owned bullets and enemy owned bullets

### DIFF
--- a/Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletTypes.cs
+++ b/Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletTypes.cs
@@ -5,4 +5,10 @@ namespace Perigon.Weapons
         NoPhysics,
         Physics
     }
+
+    public enum BulletOwner
+    {
+        Player,
+        Enemy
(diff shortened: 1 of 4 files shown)


Assets/_BForBoss/_Weapons/Scripts/Bullets/BulletTypes.cs
```diff
@@ -5,4 +5,10 @@ public enum BulletTypes
         NoPhysics,
         Physics
     }
+
+    public enum BulletOwner
```
m4rtint, line 9: so instead of having another source of Player/enemy - since we have layers and tags for players already.

we can have 2 separate bullet spawners - one for enemies and another for players. I would say it's more organized and cleaner that way.

PR #343, opened 2023-03-23
BFB-500 : Ring System

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-500

**Description**
Create Ring system based off of labor system. Currently allows for two types of ring systems - All at once, where each ring is active at the same time and they can be triggered in any order, or ordered, where each ring is active sequentially. Ordered systems can also be randomised. Currently in the sandbox scene, there is an ordered 5 ring system, an all at once 2 ring system, and then a random 5 ring system. 

**Screenshots / GIFs**
image
image


Assets/_BForBoss/_Utility/Scripts/PlayerTriggerHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace BForBoss.Utility
+{
+    public class PlayerTriggerHelper : MonoBehaviour
+    {
+        public event Action PlayerEnteredTrigger;
+        
+        private void OnTriggerEnter(Collider other)
+        {
+            if(other.gameObject.layer == 11)
```
m4rtint, line 12: what is layer 11? 

can you use TagsAndLayer class ?

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
```diff
@@ -97,6 +99,11 @@ protected virtual void Reset()
             {
                 _environmentManager.Reset();
             }
+
+            if (_ringLaborManager != null)
```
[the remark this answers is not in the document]
unoctium1: Sure, should I remove the null checks then if I'm moving it into the sandbox manager?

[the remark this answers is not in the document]
m4rtint: remove the null checks here. 

But do use a panic Helper in sandbox

unoctium1:
> @Gursaanj do we need 3 separate systems?
> 
> If we do, that's fine. But if we only are picking one for the final, then we'll have a lot of redundancy causing confusion when looking through code.

Thats a fair point, but it is really only two systems. Random and ordered are functionally the same

Assets/_BForBoss/_Core/Scripts/Managers/BaseWorldManager.cs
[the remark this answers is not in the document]
unoctium1: Sounds good, will do

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/OrderedRingSystem.cs
```diff
@@ -0,0 +1,48 @@
...
+using UnityEngine;
+
+namespace BForBoss
+{
+    public class OrderedRingSystem : RingSystem
+    {
+        private readonly bool _randomise;
+        private Queue<RingBehaviour> _ringQueue;
+
+        public OrderedRingSystem(RingBehaviour[] rings, bool randomise = false) : base(rings)
```
m4rtint, line 14: Randomize typo

m4rtint: and because it's a bool - 
_isRandomized

Assets/_BForBoss/_Utility/Scripts/PlayerTriggerHelper.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using Perigon.Utility;
+using UnityEngine;
+
+namespace BForBoss.Utility
+{
+    public class PlayerTriggerHelper : MonoBehaviour
```
m4rtint, line 7: Can call it `PlayerTriggerBehaviour` 

Cuz you can use this with a lot of other things. 

And instead of using an event, can you use a lambda instead? So you don't need to unsub

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class RingBehaviour : MonoBehaviour
+    {
+        public event Action<RingBehaviour> RingActivated;
+        
+        [SerializeField] private TMPro.TMP_Text[] _labels;
+        [SerializeField] private string _label;
+
+        private PlayerTriggerHelper _trigger;
+
+        public string Label => _label;
```
m4rtint, line 17: I don't think this was used.

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
```diff
@@ -0,0 +1,54 @@
...
+using UnityEngine;
+
+namespace BForBoss
+{
+    public class RingBehaviour : MonoBehaviour
+    {
+        public event Action<RingBehaviour> RingActivated;
+        
+        [SerializeField] private TMPro.TMP_Text[] _labels;
+        [SerializeField] private string _label;
```
m4rtint, line 13: If we have a tmp text, i don't think we need this.

Assets/_BForBoss/_Utility/Scripts/PlayerTriggerHelper.cs
unoctium1: I'm not sure I get what you mean, why wouldn't I need to unsub?

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
unoctium1: This was a convenience thing, just used by the on validate. Since its annoying to have to set the same text in multiple places

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
unoctium1: Good catch, forgot to get rid of that

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using BForBoss.Labor;
+using UnityEngine;
+using UnityEngine.Serialization;
+using Debug = UnityEngine.Debug;
+
+namespace BForBoss
+{
+    public class RingLaborManager : MonoBehaviour
```
m4rtint, line 11: Don't need to change this for now. 

But in the future, This should be just named "LaborSystem" and everything below be named "Labor" related names. 

The point of this Labor is that it is generic and able to contain anything `Labor`.  So if we visually change the Rings to Lasers, it'll still be applicable. 
If we change the Rings into buttons, it'll still be applicable. 

The higher up the hierachy we go, the more abstract the managers should be, since the components lower down (the children) will be the more specific interchangeable components.

unoctium1: Yeah I get that. This class has kinda a weird double purpose that I wasn't quite sure how best to address, where its serving both as a controller for the labor system and a view for setting up rings. Ideally I think those should be broken up

Assets/_BForBoss/_Core/Scripts/Managers/SandboxWorldManager.cs
```diff
@@ -1,11 +1,37 @@
...
+            base.Start();
+            _ringLaborManager.Initialize();
+        }
+
+        protected override void OnValidate()
+        {
+            base.OnValidate();
+            if (_ringLaborManager == null)
+            {
+                PanicHelper.Panic(new Exception("_ringLaborManager is missing from World Manager"));
```
m4rtint, line 33: SandboxWorld Manager

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
m4rtint: What's the text you have to set?

unoctium1: The number on the ring, its two different tmp objects with the same label

m4rtint: Okay i see, 
Then you can probably just `FindChildComponent` for the TMP_Text.

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
m4rtint: We'll work with what we have here for now as the first iteration.

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
unoctium1: Good call yeah

Assets/_BForBoss/_Utility/Scripts/PlayerTriggerHelper.cs
unoctium1: Wait so you just mean keep it as an action but remove the event modifier?

m4rtint: Yeah

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -0,0 +1,77 @@
...
+            foreach (var grouping in _systemsToBuild)
+            {
+                RingSystem newSystem = grouping.RingSystemType switch
+                {
+                    RingSystemTypes.Standard => new OrderedRingSystem(grouping.Rings),
+                    RingSystemTypes.DisplayAllAtOnce => new AllAtOnceRingSystem(grouping.Rings),
+                    RingSystemTypes.RandomSelection => new OrderedRingSystem(grouping.Rings, true),
+                    _ => throw new ArgumentOutOfRangeException()
+                };
+                newSystem.OnLaborCompleted += () => Debug.Log($"Completed {grouping.Rings.Length} ring {grouping.RingSystemType} system");
```
Gursaanj, line 50: Nitpick: Could use PerigonLogger, with Key "Labor"

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -0,0 +1,77 @@
...
+                _ringSystems.Add(newSystem);
+            }
+        }
+
+        private void Update()
+        {
+            if(_laborSystem == null) return;
+            if (!_hasCompletedSystem && _laborSystem.IsComplete)
+            {
+                Debug.Log("All labors completed");
```
Gursaanj, line 60: same here with Logger

Gursaanj (approved):
Looks great!! I do agree with Martin that maybe just the 2 systems instead of all 3 is good enough! Now the next step is looking into connecting this to setting/resetting the Timer

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/AllAtOnceRingSystem.cs
```diff
@@ -0,0 +1,40 @@
...
+        
+        public override void Activate()
+        {
+            foreach (var ring in _remainingRings)
+            {
+                ActivateRing(ring);
+            }
+        }
+
+        protected override void SetupRingLists(IList<RingBehaviour> rings)
```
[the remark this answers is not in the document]
unoctium1: Yeah I get that, just wasn't sure the best way to deal with it

PR #349, opened 2023-03-27
Add Debug Functionality to EnergySystem

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-477

**Description**
Gave the players the ability to freeze costs (i.e. make everything cost 0 energy). They can technically have infinite energy when needed but can then toggle back to the same amount of energy they had earlier when desired. 

It also made a quick and easy method to allow players to give themselves a full amount of energy. To make testing easier

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
@@ -122,7 +122,7 @@ namespace BForBoss
                 {
                     GUILayout.Space(0.15f * _windowRect.height);
                     _playerLifeCycle.IsInvincible = DrawDebugToggle(_playerLifeCycle.IsInvincible, new GUIContent("Player Invincibility"));
-                    _energySystemBehaviour.UseDebugEnergyCost = DrawDebugToggle(_energySystemBehaviour.UseDebugEnergyCost, new GUIContent("Freeze Energy Cost"));
+                    _energySystemBehaviour.UseDebugEnergySystemConfig = DrawDebugToggle(_energySystemBehaviour.UseDebugEnergySystemConfig, new GUIContent("Freeze Energy"));
 
                     if (GUILayout.Button("Give Max Energy"))
                     {
--- a/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
@@ -18,6 +18,7 @@ namespace BForBoss
         [SerializeField] private FreeRoamCamera _freeRoamCamera = null;
         
         private PlayerLifeCycleBehaviour _playerLifeCycle;
+        private EnergySystemBehaviour _energySystemBehaviour;
         private bool _isPlayerCurrentlyInvincible;
         
         //State Handling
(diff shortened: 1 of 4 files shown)


Assets/_BForBoss/_Core/Scripts/EnergySystem/EnergySystemBehaviour.cs
```diff
@@ -179,6 +179,13 @@ private float MapToAccrueValue(EnergyAccruementType type, float multiplier)
         
         private float MapToExpendValue(EnergyExpenseType type, float multiplier)
         {
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+            if (UseDebugEnergyCost)
```
m4rtint, line 183: We shouldn't be leaking Debug logic into the implementation itself. We want to keep it as separate as possible. 

If you want to make it so that the costs are all 0. Then create a new
`_energySystemConfigurationData = new GODMODECONFIG()`, and then set your own values in the partial. 

And then you can revert it after with the same `_energySystemConfigurationData = _energySystemConfiguration.MapToData();`. 

The point of setting everything as Data is that you can edit just the data/state, and it should work as you would like it to.

m4rtint (changes requested):
Should change it so you set the `data` instead of changing the logic.

Assets/_BForBoss/_Core/Scripts/EnergySystem/EnergySystemBehaviour.Debug.cs
```diff
@@ -0,0 +1,14 @@
+#if UNITY_EDITOR || DEVELOPMENT_BUILD
+namespace BForBoss
+{
+    public partial class EnergySystemBehaviour
+    {
+        public bool UseDebugEnergyCost = false;
+
+        public void SetMaxEnergy()
+        {
+            EnergyData = _energyData.Apply(_energyData.MaxEnergyValue);
```
m4rtint, line 10: you should write which parameter your setting

Gursaanj: Updated

Assets/_BForBoss/_Core/Scripts/Debug/DebugWindow.cs
```diff
@@ -120,6 +122,12 @@ private void CreateDebugActionList()
                 {
                     GUILayout.Space(0.15f * _windowRect.height);
                     _playerLifeCycle.IsInvincible = DrawDebugToggle(_playerLifeCycle.IsInvincible, new GUIContent("Player Invincibility"));
+                    _energySystemBehaviour.UseDebugEnergySystemConfig = DrawDebugToggle(_energySystemBehaviour.UseDebugEnergySystemConfig, new GUIContent("Freeze Energy"));
+
+                    if (GUILayout.Button("Give Max Energy"))
+                    {
+                        _energySystemBehaviour.SetMaxEnergy();
```
m4rtint, line 129: Is there a point to this if you're going to freeze energy anyway?

Gursaanj: Yup. In the case of testing which values feel the best for costs (i.e. shooting and slow-mo) its nice to be able to test how they feel over and over again without needing to manually accrue the energy.

m4rtint (approved):

PR #350, opened 2023-03-27
BFB-525: Make utility logger robust to dark theme

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-525

**Description**
Quick fix for a bug I found while setting up the labor system debugs. Changes the default option for logging to a setting that varies based on user theme. In dark theme right now, the old default of black text was unreadable


unoctium1:
One alternative, probably better fix for this would just be not to colorize the string as the default behaviour

m4rtint (approved):

PR #351, opened 2023-03-30
BFB-500: Refactor Ring Behaviour

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-500

**Description**
Change all at once behaviour to use a queue and consolidate into one script (cause all the behaviour is about the same now anyways)


Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -42,9 +42,9 @@ private void CreateSystems()
             {
                 RingSystem newSystem = grouping.RingSystemType switch
                 {
-                    RingSystemTypes.Standard => new OrderedRingSystem(grouping.Rings),
-                    RingSystemTypes.DisplayAllAtOnce => new AllAtOnceRingSystem(grouping.Rings),
-                    RingSystemTypes.RandomSelection => new OrderedRingSystem(grouping.Rings, true),
+                    RingSystemTypes.Standard => new RingSystem(grouping.Rings),
+                    RingSystemTypes.DisplayAllAtOnce => new RingSystem(grouping.Rings, allAtOnce: true),
+                    RingSystemTypes.RandomSelection => new RingSystem(grouping.Rings, isRandomized: true),
```
m4rtint, line 47: @Gursaanj are we still doing random? 
or are we doing predesigned only?

m4rtint:
So this change makes it so 1 ring shows up one at a time. 
Are we not doing all at the same time + billboarding numbers? 

or is that for another ticket? 

@Gursaanj

unoctium1:
@m4rtint I only changed the behaviour of the all at once ring system. Now they all show up at once still, but must be completed in order. Standard and random are unchanged

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
Gursaanj: We will be doing preassigned but I say lets just keep this in for now, just so we can pla

> So this change makes it so 1 ring shows up one at a time. Are we not doing all at the same time + billboarding numbers?
> 
> or is that for another ticket?
> 
> @Gursaanj

The Billboarding will be another ticket!

Gursaanj: We are only doing preassigned, so I think it is safe to delete the random selection

unoctium1: I removed the random selection from the scene - Now the current demo loop is 5 all at once, 2 all at once, and then 5 in order sequential

m4rtint:
> @m4rtint I only changed the behaviour of the all at once ring system. Now they all show up at once still, but must be completed in order. Standard and random are unchanged

i meant like remove the randomized code. 

But that's fine for now.

m4rtint (approved):

PR #360, opened 2023-04-20
BFB-527 & BFB-501 - Integrate Countdown Timer and Ring System

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-501
https://perigongames.atlassian.net/browse/BFB-527

**Description**
Integrate rings with countdown timer. Each ring system now has a specified duration and will reset if not complete when the duration ends. Also add some functionality to countdown timer


Assets/_BForBoss/_Core/Scripts/HUD/CountdownViewBehaviour.cs
```diff
@@ -12,19 +12,36 @@ public class CountdownViewBehaviour : MonoBehaviour
         private bool _isRunning;
 
         private float _amountOfTime;
+        
+        private Action OnCountdownCompleted;
+        
+        public bool IsRunning => _isRunning;
 
-        public void StartCountdown(float amountOfTime)
+        public void StartCountdown(float amountOfTime, Action onCountdownCompleted = null)
```
m4rtint, line 20: Imo, the count down HUD / UI, shouldn't be the one that determines it ended. 

The logic of "The time has ended, now do this" should be contained within the Manager itself. 
This UI/HUD behaviour should soley be in charge of letting users know the time.

unoctium1: Yeah that was mostly a result of how CountdownViewBehaviour currently is setup, I can rewrite it to be more of an actual view though

m4rtint: I just remembered countDownViewBehaviour is more than not a placeholder. This part is flexible

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -29,22 +27,31 @@ public void Reset()
             _laborSystem = new LaborSystem(_ringSystems);
         }
         
-        public void Initialize()
+        public void Initialize(CountdownViewBehaviour countdownView)
```
m4rtint, line 30: `CountdownViewBehaviour` 

Make this a singleton, or just call its static shared instance.

Gursaanj (approved):

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
unoctium1: Split countdown view into a singleton countdown timer, and turned the view into more of a view for that

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -35,16 +32,24 @@ public void Initialize()
             _laborSystem = new LaborSystem(_ringSystems);
         }
 
+        public void ToggleTimer()
+        {
+            if(CountdownTimer.Instance.IsRunning)
```
m4rtint, line 37: if you had to call this many `CountdownTimer.instance`, you can probably put this whole method intothe `countdownTimer`itself.

unoctium1: Yeah good call, the toggle functionality will be a lot cleaner inside it as well

m4rtint (approved):

PR #361, opened 2023-04-22
BFB-532: Missile and Boss Shields Collision ignore

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-532

**Description**
- Ignore Physiccs

**Screenshots / GIFs**

https://user-images.githubusercontent.com/5779541/233790065-a20b61bf-0aa2-42d2-b8a9-28a5bd9b6e44.mp4


Assets/_BForBoss/_Core/Scripts/Boss/DerekMissileBehaviour.cs
```diff
@@ -151,5 +153,25 @@ private void OnEnable()
             _elapsedAutoPilotTimeToLive = 0;
             _state = State.TowardsApex;
         }
+
+        private void Start()
+        {
+            //NOTE - *Physics.IgnoreCollision*
+            //Note that this code should be run in the Start() method or later, since the colliders need to be initialized before you can use them. 
+            var missileCollider = GetComponent<Collider>();
+            var deathAreas = FindObjectsOfType<DeathAreaBehaviour>()
```
Gursaanj, line 162: Do we want to make sure we handle the Inactive `DeathAreaBehaviors` as well, since not all of them will active at the same time

I think its like `FindObjectsOfType<DeathAreaBehavior>(includeInctive: true)`

Gursaanj (approved):

Assets/_BForBoss/_Core/Scripts/Boss/DerekMissileBehaviour.cs
m4rtint: oh yeah you're right.

PR #362, opened 2023-04-26
BFB 515: Derek Context Manager WIP

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-515

**Description**
First Draft for managing the encounter context for the boss (Derek). This Context manager will shift the boss's phases and conditions based on the environment's context.

i.e. Reaching a certain health level advances the phase, and thus shifts the Vulnerability of the boss
i.e. Completing a ring labor in time will switch the vulnerability
i.e. Pressing a button at the beginning will shift the encounter to the first phase

It is a simple listen-and-call model and will most likely need to change to be more flexible. **SO JUST PUTTING THIS HERE FOR NOW TO START THE DISCUSSION**

my idea is that the world manager will initialize the Context manager who will, in turn, initialize and claim responsibility for the Boss Manager (in charge of moving and attacking the player) and the RingLaborManager (in charge of setting up the ring labors)


Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -0,0 +1,175 @@
...
+                case Phase.SecondPhase:
+                    _currentPhase = Phase.FinalPhase;
+                    break;
+                case Phase.Death:
+                    return;
+                default:
+                    break;
+            }
+
+            _ringLaborManager.Initialize(OnLaborCompleted);
```
m4rtint, line 160: The method name `Initialize` implies that you should only be calling this method once in the beginning. But it's placed within a method that is called multiple times?

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -0,0 +1,175 @@
...
+            //TutorialEndButton.Bind(this, OnPhaseChanged); 
+            _LaborIncompleteWaitForSeconds = new WaitForSeconds(_LaborIncompleteTimer);
+        }
+        
+        private void OnLaborCompleted(bool wasSuccessful)
+        {
+            if (_currentPhase == Phase.Tutorial || _currentVulnerability != Vulnerability.Invulnerable)
+            {
+                Perigon.Utility.Logger.LogError("Labor should not have have been active during the tutorial phase, or when the boss is not invulnerable. Exiting Encounter");
+                return;
```
m4rtint, line 127: So if there's an error, is the boss just going to be stuck? how should the game handle it?  

-  Could reset the game? 
- crash the game? 
- Load with default settings?

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace BForBoss
+{
+    public class DerekBossManager : BossManager
+    {
+        private Action _onPhaseComplete;
+        
+        public void Initialize(Action OnPhaseComplete)
+        {
+            _onPhaseComplete = _onPhaseComplete;
```
m4rtint, line 12: typo

Assets/_BForBoss/_Core/Scripts/Managers/ContextManager.cs
```diff
@@ -0,0 +1,11 @@
+using UnityEngine;
+
+namespace BForBoss
+{
+    /// <summary>
+    /// Base manager for handling and managing all context between Bosses, their labors and the environment
+    /// </summary>
+    public abstract class ContextManager : MonoBehaviour
```
m4rtint, line 8: imo, don't need a parent abstract class just yet. Unless you already have a clear vision of what the other boss behaviours are like and have a plan for this class.  

i rather have 3+ of the same behaviours and then abstracting out, than trying to predict its usage - leading to more tech debt if it's not used.

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Collections;
+using Sirenix.OdinInspector;
+using UnityEngine;
+
+namespace BForBoss
+{
+    public class DerekContextManager : ContextManager
+    {
+        [SerializeField, Tooltip("How long should the system wait before restarting the Labor after an incomplete Labor")]
+        private float _LaborIncompleteTimer = 5.0f;
```
m4rtint, line 11: typo -  _laborIn...

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
```diff
@@ -26,8 +28,9 @@ public void Reset()
             _laborSystem = new LaborSystem(_ringSystems);
         }
         
-        public void Initialize()
+        public void Initialize(Action<bool> onLaborCompleted)
```
m4rtint, line 31: just a btw, you don't have you put it in the init if u always update it.

you can just make it a `public Action<Bool>` and update it whenever needed.

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -0,0 +1,175 @@
...
+            debug_dbm.UpdateVulnerability(debug_currentVulnerability);
+        }
+#endif
+        
+        
+        public void Reset()
+        {
+            if (_ringLaborManager == null || _bossManager == null)
+            {
+                return;
```
m4rtint, line 104: SInce this class requires `_ringLaborManager` and `_bossManager` . you can use the Panichelper here and crash the game.

And then remove all the `null` checks here.

We don't want to suppress/ignore the problem, we want to be able to find it asap and handle it.

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
unoctium1, line 11: Right now, the labors (or specifically the ring system, I didn't pull the behaviour out to the general labors) already handle this themselves. I also kinda think that makes more sense, cause that way we can have time as a function of the number of rings or something like that. You'd be better of passing an event up through the ring system and listening for that rather than separately managing time here

m4rtint: As long as it's
1. single source of truth.
2. makes sense within the context of the class.

Assets/_BForBoss/_Core/Scripts/Managers/ContextManager.cs
m4rtint: This is more of an opinion. It's up to you.

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
unoctium1: Seconding this, I think you should be using `Reset` here and pulling your `OnLaborCompleted` out into its own method. `RingLaborManager.CreateSystems` is really only intended to be called once

Assets/_BForBoss/_Core/Scripts/Managers/ContextManager.cs
Gursaanj: Good point, Initially I had an idea of creating some parental logic but couldn't justify it in the end, will remove it!

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
Gursaanj: good catch! At work, Unreal coding convention uses PascalCase for variables instead of camelCase, so my brain has been kinda scrambled haha

Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingLaborManager.cs
Gursaanj: Actually, I didn't have to keep updating `onLaborCompleted` and I only need to initialize the RingLabor once, right as we leave the Tutorial phase and enter the First Phase. Every phase change after, we just need to reset the RingLabor system

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
Gursaanj: Sounds good! I like that better!

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
Gursaanj: Would it be better to have it in the `Initialize` method then, rather than on `Reset`?

m4rtint: Wherever it gets initialized once, you can put it there.

Gursaanj:
Updated to only call ring labor manager initialize once we get into the first phase!

In reality, we would want DerekContextManager in charge of initializing and resetting the RingLaborManager and not the World Manager but left it as is until the systems are in place to start changing phases (next PR)

Also added the DerekContextManager and DerekBossManager to the FullSandbox Scene

unoctium1 (approved):
Good to merge in, following the talk after this past meeting!

PR #363, opened 2023-04-30
BFB-535: Fix wall of death only working from one side 

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-535

**Description**
Previously, wall of death colliders worked when still but while in motion, they only registered collisions from the front side and not when the player ran into the back. This was fixed by adding rigidbodies both to the walls themselves and to the parent column, and then modifying the rotation script so that if a rigidbody was present it will use the rigidbody rotation rather than transform rotation.

I also replaced the wall of death colliders with box colliders, ended up not being related to the issue but figured its a worthwhile change to keep


Assets/_BForBoss/_Core/Scripts/Boss/RotationalMovementBehaviour.cs
```diff
@@ -31,12 +39,21 @@ public void StopRotation()
             _direction = 0;
         }
 
-        private void FixedUpdate()
+        private void Update()
         {
-            if (_direction != 0)
+            if (_direction != 0 && _rigidbody == null)
             {
                 transform.Rotate(Vector3.up, _direction * _rotationRate * Time.fixedDeltaTime);
```
Gursaanj, line 46: This should be `Time.DeltaTime` instead now

Gursaanj (approved):
Looks good. One small comment to address

PR #364, opened 2023-05-10
BFB-542: Add nested ring sequences

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-542

**Description**
Adds a new type of ring labor, for nested ring systems. This allows a series of systems to be treated as one labor, with a fixed time. Nested systems are always randomized (currently) but their subsystems can be random, all in order, all at once, or any combination of the three. Nested systems have one 'master' time and their subsystems have no control over time

**Screenshots / GIFs**
image
In this image, two subsequences of two rings each have been added, and the user has 20 seconds to complete both sequences. The second sequence has a time of 5 in the inspector, but this is ignored at runtime


Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/NestedRingSystem.cs
```diff
@@ -0,0 +1,95 @@
...
+        private void SetupSystems(IList<RingSystem> systems)
+        {
+            if (_isRandomized)
+            {
+                systems.ShuffleFisherYates();
+            }
+            _queue = new Queue<RingSystem>(systems);
+        }
+
+        private void CountdownFinish()
```
Gursaanj, line 63: private methods should be after protected ones

Gursaanj (approved):
Small comment, but looks great!

PR #368, opened 2023-05-31
BFB-549: Update labor system to work better with context manager

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-549

**Description**
Labor system is now fully controlled by the context manager. Note that this has some caveats: The rings will not appear until after the first phase is started (using the debug button on the context manager)
Initialization & starting are now different operations
Resetting no longer re activates the system. Note that resetting now soft locks the game, since the context manager only starts the ring system on phase 1 and resetting doesn't reset the current phase
Ring labels are set dynamically, within each ring system. Each system will start with ring 1.


Gursaanj (approved):
Looks great!, Could you add a gif/video of the changes onto the Jira ticket

PR #371, opened 2023-06-21
BFB-541: Launch Player from Shield

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-541

**Description**
Set up derek shield behavior to launch the player with configurable settings for launch.
Tested against both just walking into the shield and enabling the shield when the player is inside it. If the player walks into it, their momentum will affect launch velocity. Otherwise, a minimum value will be used.


Gursaanj (approved):
Looks great, could you attach a video/gif to the Jira card for visualization when you have the chance!

PR #372, opened 2023-06-27
BFB-550 : Create working flow for Boss encounter

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-550

**Description**
This creates the specific flow we should expect from our encounters

- Start by shooting the tutorial button -> Initiating phase 1 Invulnerable state
- Do Ring labor while dodging missiles and Death Wall ->Succeed to set boss to being vulnerable
- While Vulnerable shoot the boss (boss has 1 health in each phase for this PR) to end the phase /OR wait ten seconds for the boss to go back into its invlunerable state again and reset the Ringlabor
- When Phase 1 ends -> Phase 2 begins (which is currently the exact same as phase 1 for the purpose of this PR)
- When Phase 2 ends -> Phase 3 begins (which is currently the exact same as phase 1 for the purpose of this PR)
- When Phase 3 ends -> A log statement appears stating that you have won!

Things to note for this PR:

- There are no differences between phases in this PR (that will be handled in the subsequent PRs) - this was made purely to establish the flow
- Nothing special happens when the game ends besides the log. Nothing to expect
- **THE GAME STATE BREAKS WHEN YOU DIE  DURING YOUR ENCOUNTER. This will be fixed once we fix our Reset logic**
- **THE GAME STATE BREAKS IF YOU DON'T SOLVE THE RING LABOR IN THE ALLOTTED TIME. This will be fixed once we fix our Reset logic for the Ring Labor**

If the game state breaks, quit the application and re-open it.

This can only be tested in the `DerekFlowSandbox.unity` Scene at the moment. added to EditorBuildSettings so that it can be tested in the build.

**Screenshots / GIFs**
https://github.com/PerigonGames/BForBoss/assets/23490604/bb51563d-25d2-49a2-9f91-99e3a3063b69


Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
```diff
@@ -1,24 +1,50 @@
...
+            }
+            _shieldBehaviour.ToggleShield(false);
+        }
+
+        public void Initialize(PlayerMovementBehaviour playerMovementBehaviour, Action<bool> onVulnerabilityExpired)
+        {
+            _wipeoutWallsManager.Initialize(playerMovementBehaviour);
+            foreach (DerekMissileLauncherBehaviour missileLauncher in _missileLauncherBehaviours)
+            {
+                if (missileLauncher == null)
```
m4rtint, line 39: When would missileLauncher ever be null? 
If it's because of a design mistake, then you should check if it's null onawake and PanicHelper crash the game. Instead of needing to check each time you loop through.

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
```diff
@@ -31,55 +57,102 @@ public void UpdatePhase(DerekContextManager.Phase phase)
...
-                    Debug.Log("Raising Shields");
-                    Debug.Log("Resetting Position");
-                    Debug.Log("Activating Death Wall Closest to player");
-                    Debug.Log("Rotating to make sure death wall gets closer to player");
-                    Debug.Log("Starting Homing Missiles");
-                    Debug.Log("------------------------");
-                    
+                    _isVulnerable = false;
+                    _shieldBehaviour.ToggleShield(true);
+                    foreach (DerekMissileLauncherBehaviour missileLauncher in _missileLauncherBehaviours)
```
m4rtint, line 88: should check the null case in onawake. Since now, it'll be inconsistent here

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
```diff
@@ -31,55 +57,102 @@ public void UpdatePhase(DerekContextManager.Phase phase)
...
+            if (_shieldBehaviour.IsActive())
+            {
+                Perigon.Utility.Logger.LogWarning("Boss was shot even with shields up, which shouldn't be possible, ignoring damage", LoggerColor.Yellow, "derekboss");
+                return;
+            }
+            
+            _onVulnerabilityExpired?.Invoke(true);
+        }
+
+        private void OnValidate()
```
m4rtint, line 123: I personally think we should use OnAwake for ingame configuration since this function is only called within the unity editor and not the executable game. 
https://github.com/Demigiant/dotween/issues/539

So if there's an issue, and we build it, the build itself won't call this.

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -9,8 +9,8 @@ namespace BForBoss
     public class DerekContextManager : MonoBehaviour
```
m4rtint, line 9: IMO, DerekContextManager cs DerekBossManager. If you ask someone what's the difference without explaining to them it's a bit confusing.

What if you called it DerekFiniteStateMachine?

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
```diff
@@ -33,38 +32,28 @@ public enum Vulnerability
         
         public void Reset()
         {
-            _ringLaborManager.Reset();
             _bossManager.Reset();
             _endTutorialButton.Reset();
         }
 
-        public void Initialize(RingLaborManager ringLaborManager, DerekBossManager bossManager)
+        public void Initialize(RingLaborManager ringLaborManager, PlayerMovementBehaviour playerMovementBehaviour)
```
m4rtint, line 39: I don't think you need `PlayerMovementBehaviour` here, but you need `IGetPlayerTransform`

The context of what you need matters here.

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
```diff
@@ -31,55 +57,102 @@ public void UpdatePhase(DerekContextManager.Phase phase)
...
-                    //Ensure Derek is invincible in this case (edge case)
-                    
-                    Debug.Log("Raising Shields");
-                    Debug.Log("Resetting Position");
-                    Debug.Log("Activating Death Wall Closest to player");
-                    Debug.Log("Rotating to make sure death wall gets closer to player");
-                    Debug.Log("Starting Homing Missiles");
-                    Debug.Log("------------------------");
-                    
+                    _isVulnerable = false;
```
m4rtint, line 86: so you have an `isInvulnerable` flag. but then also have the `Vulnerability` State. 

Can you only store one state? instead of having 2 sources of truth? `DerekContextManager.Vulnerability.Invulnerable` vs isVulnerable

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
Gursaanj: Good point

Assets/_BForBoss/_Core/Scripts/Boss/DerekBossManager.cs
Gursaanj: Good point, changed the var to be `DerekContextManager.Vulnerability` instead of a bool. It should be noted that this var is only being used at the moment for the temp Damage Receiving logic I have in the update loop, and will be removed as soon as the HealthBehaviour is implemented!

Added the todo comments in the cs to better clarify that

Assets/_BForBoss/_Core/Scripts/Boss/DerekContextManager.cs
Gursaanj: Yeah, so the way I have reasoned is the `ContextManager` is in charge of the current state of the encounter, Handle responding to the changes in the Labor/s (i.e. Ring labor) and Handle responding changes of the Boss.

So like if, for example, when the player completes a ring labor while Derek is phase 3 (the final phase) we want to change the lighting of the scene and play a different soundtrack, it would be the Context Manager who would be calling that. 

Thats why I am not so keen calling this class a FiniteStateMachine, even though that it what it basically is at the moment.

Gursaanj: I am down to change the name though! Here are some suggestions

`DerekDirector` (Director referring to the fact that this manager is basically in charge of all the main "Actors" in the scene and make them come together) - got this idea from a project I previously worked on
`DerekEnvirornmentManager` - A bit more on the nose


Down for other ideas

m4rtint: Personal preference,i don't mind if u stick with context.

m4rtint (approved):

PR #379, opened 2023-07-24
BFB-561: Tutorial section in settings

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-561

**Description**
- Tutorial section added for settings
- Resized Setting screen

**Screenshots / GIFs**

https://github.com/PerigonGames/BForBoss/assets/5779541/4090ceb9-57f8-4890-b7cd-220ef9d31c6e


Assets/_BForBoss/_Core/Scripts/UserInterface/Tutorial/TutorialCanvas.cs
```diff
@@ -0,0 +1,89 @@
...
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace BForBoss
+{
+    public class TutorialCanvas : MonoBehaviour
+    {
+        [Title("Buttons")] 
+        [SerializeField] private Button _backButton;
+        [SerializeField] private Button _basicButton;
```
Gursaanj, line 14: It could easier for scaling to use a dictionary (and use Odin serialize it) of `<Button, CaroselView>` for all of the tutorials. So that we can just add more tutorials in the inspector in the future rather than copy/paste in the code.

And iterate over the values for the `ForEachTutorialCarouselView` method

m4rtint: Because it's only like 4 - 5, was thinking it was fine.

Don't really wanted to add more if it's not going to change too much.

Gursaanj (approved):

PR #383, opened 2023-08-01
BFB-556 - Model Integration

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-556

**Description**
Set up Derek V2 model, colliders, and animations. Also replace placeholder rocket prefabs with models. Note that some scripts had to be changed to support compound colliders and layer masks. Also changed wall behavior to deactivate gameobject when collapsed, cause the collapsed walls looked pretty bad otherwise (they're visible in the recording). Everything looks pretty good, I do think we might want to replace the material on Derek's head cause it looks a little odd with the death walls going through it (should be pretty straightforward with a custom shader or some material settings though)

**Screenshots / GIFs**
https://youtu.be/G_AlJwDbjpk

Different Head Material Settings: 
Default - 
image
Depth Prepass - 
image


m4rtint:
Looks good. 
but want @Gursaanj to confirm from the video, that this is what he imagined

Gursaanj (approved):
Looks great, will go ahead and create a ticket for the material task during today's meeting

PR #390, opened 2023-08-16
BFB-572 - Ring System Color Changing

unoctium1:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-572

**Description**
Change ring color with every labor. Also expose the main color of the shader and set up decent looking colors in the DerekFlowSandbox scene

### DIFF
--- a/Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
+++ b/Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
@@ -1,5 +1,6 @@
 using System;
 using BForBoss.Utility;
+using Perigon.Utility;
 using UnityEngine;
 
 namespace BForBoss.RingSystem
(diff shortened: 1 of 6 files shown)


Assets/_BForBoss/_Core/Scripts/Labor/RingSystem/RingBehaviour.cs
```diff
@@ -9,10 +9,22 @@ public class RingBehaviour : MonoBehaviour
...
         private PlayerTriggerBehaviour _trigger;
 
+        public void SetLabel(string label)
+        {
+            _labelTMP.text = label;
+        }
+
+        public void SetColor(Color color)
+        {
+            if (_renderer == null) return;
```
m4rtint, line 24: instead of doing null check every time, Do a Panic to check for the renderer on Awake

m4rtint (approved):

PR #411, opened 2023-09-06
BFB-612: More variance in platform shapes

csweet22:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-612

**Description**
More grey platforms have more interested shapes.

Please take a look and see if there is an area that you think could use more variation.


m4rtint:
This is a great first pass, the variable shapes makes for a more live environment. 

But I was hoping for more height variability like this. 
image

m4rtint (approved):

PR #414, opened 2023-09-14
BFB-614: Missile COnfigurations

m4rtint:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-614

**Description**
- Changed - Height of Apex (So it's higher)
- Changed Damage to 10
- Time towards apex - Time taken to get to apex is slowed to 3 seconds
- Knock back down to 40.

Not sure if there's any other configurations you want to make, 

**Screenshots / GIFs**
image


m4rtint:
Should honestly test it out

Gursaanj (approved):
Definitely feels better, still feels like the missile doesn't hit all the time if I stand still on some of the higher scaffolding. 

Do you know if the missile aims towards the camera, or the feet of the player?

m4rtint:
> Definitely feels better, still feels like the missile doesn't hit all the time if I stand still on some of the higher scaffolding.
> 
> Do you know if the missile aims towards the camera, or the feet of the player?

In Theory, It should be from the player center of Grav. could try and force it to be higher on the chest

PR #416, opened 2023-09-20
BFB-616: Edit First Phase Ring Systems

Gursaanj:
**JIRA Ticket**
https://perigongames.atlassian.net/browse/BFB-616

**Description**
Edited the ring systems we used for the first phase to get general variety in what locations the tutorials take place

### Big Change
From playing it a couple of times, I really like the idea of the first-ring system (Ring_Ring_1, Ring_2) being consistent between runs. for 2 reasons

- For first-time players, it is super helpful to have the first ring pop up right next to them when starting the Boss battle (It can seem super confusing if the first ring is really far away)
- Creates a nice flow into gameplay when restarting the level after dieing (like a nice pattern to get into the middle of gameplay)

To achieve this, I created a Separate RingGrouping for Phase 1, to start with this consistent Ring system (Ring, Ring_1, Ring_2) and after the player has achieved that, the actual Ring grouping (of 3 possible ring systems - image below) is used for the remainder for phase 1.

Let me know if this makes sense and if it sounds desirable. If it doesn't, I will revert my changes and just use the new system (as seen in the image below)
**Screenshots / GIFs**
Screenshot 2023-09-20 024752


m4rtint:
Does this mean that if you get the health down to phase 2, it'll skip phase 1? 

PHase 1 tutorial -> Phase 2?

m4rtint (approved):

Gursaanj:
> Does this mean that if you get the health down to phase 2, it'll skip phase 1?
> 
> PHase 1 tutorial -> Phase 2?

Yup, if the player is able to complete phase 1 with only the first (special) ring system then yeah they will skip the other Phase 1  ring systems! Which I think is fine, personally

m4rtint (approved):

m4rtint (changes requested):
Need to make sure this is fixed

Gursaanj:
> Need to make sure this is fixed

Is it still broken? Should be fixed now

m4rtint (approved):

